PR #4, opened 2022-01-26
'Solution'

oleksandra-halytska:


app/main.py
```diff
@@ -1 +1,51 @@
...
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name):
+        super().__init__(name)
+        super().__init__(name)
+
```
alix-mak, line 47: You don't need to call super() twice, I think. Try to do it once.

alix-mak (approved):
You don't need to call super() twice, I think. Try to do it once.

app/main.py
MasterpieceElbow: Try to remove duplicated line

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #13, opened 2022-01-26
Solution

roland813:


denyslipin (approved):

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name):
+        super(FullStackDeveloper, self).__init__(name)
+
+    def create_web_application(self):
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
+        FrontendDeveloper.create_awesome_web_page(self)
```
MasterpieceElbow, line 47: It is bad style to hardcode classes here. It would be better write `self.create_powerful_api()` and let mro find appropriate method.

MasterpieceElbow (changes requested):

Krimat (approved):

MasterpieceElbow (approved):

PR #14, opened 2022-01-26
Solution

nikon4uk:


serhii56465 (approved):
I used self.skills.extend(["JavaScript", "HTML", "CSS"]) instead of learn_skill . Just in case, to you know that this way works too.

MasterpieceElbow (approved):

PR #16, opened 2022-01-26
Solution

vldtsw:


liuda-chuzdiuk (approved):

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def create_powerful_api(self):
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
MasterpieceElbow, line 38: You created self.skills in `SoftwareEngineer` and here you reassign it to another list. You lost your initial link. It would be better you to extend existed list.

MasterpieceElbow (changes requested):

app/main.py
vldtsw: Fixed it

MasterpieceElbow (approved):

PR #21, opened 2022-01-27
'Solution'

Oleksandr563487953:


app/main.py
```diff
@@ -1 +1,75 @@
...
+    def learn_skill(cls, skill):
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = [
+        "JavaScript",
+        "HTML",
+        "CSS"
+    ]
```
nikon4uk, line 17: I believe that in this task, the skills should be stored in the attributes of the constructor, not the class, since the examples assume that the instances of these classes are a specific programmer with his skills, and not the position on which he works, if you want to add or remove a skill, you remove it for all programmers, since they will refer to a common attribute, and not their own

app/main.py
```diff
@@ -1 +1,75 @@
...
+
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = [
+        "Python",
+        "SQL",
+        "Django"
+    ]
```
nikon4uk, line 33: same

app/main.py
```diff
@@ -1 +1,75 @@
...
+        print(f"{self.name} is creating an API...")
+
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    skills = [
+        "Java",
+        "Android studio"
+    ]
```
nikon4uk, line 48: same

nikon4uk (changes requested):

app/main.py
Oleksandr563487953: Yes, i agree. It was obvious to provide that list for every developer, but i didn`t notice that. Changes provided.

nikon4uk (approved):

app/main.py
```diff
@@ -1 +1,69 @@
...
+    def __init__(self, name):
+        super(FullStackDeveloper, self).__init__(name)
+        self.skills = [
+            "Python",
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS"
+        ]
```
MasterpieceElbow, line 64: So now it is bad style because you have to add skills that you have added in parents, you repeat yourself.
In every class except `SoftwareEngineer` you __reassign__ attribute `self.skills` that you assign in `SoftwareEngineer`. If you extend `self.skills` instead of __reasign__ - you won't have to assign that big list in `FullstackDeveloper` and you will have this skills just by using super().__ init__.

MasterpieceElbow (changes requested):

app/main.py
Oleksandr563487953: Thank you, changes provided. I didn`t know that FullstackDeveloper will inherit self.skills both from FrontendDeveloper and BackendDeveloper simultanously, just by super(), and self.skills will be sliced. I hope, now i understand super() a bit better.
P.S. I hope, i understood the problem correctly.

IvanRamyk (approved):

PR #32, opened 2022-05-23
Solution

YuriiKindrat:


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def learn_skill(self, skill):
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str):
+        super().__init__(name)
+        for skill in ["JavaScript", "HTML", "CSS"]:
```
Aleksei-Isaev, line 15: Here and below, maybe better use "extend"  instead of "for", but it's up to you.

Aleksei-Isaev (approved):

MasterpieceElbow (approved):

PR #36, opened 2022-05-24
'Solution'

Aleksei-Isaev:


NOSSUDAK (approved):
It would work even without lines 41, 42. Also you can add type annotation in som places

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name):
+        super().__init__(name)
```
MasterpieceElbow, line 42: You don't have to implement `__init__` here. Without that, it works the same.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #45, opened 2022-05-24
'Solution'

daria-niklaevna:


app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name):
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str):
```
SergeyDev1996, line 6: it would be way more logical to use @classmethod for 'learn skill' function. In this, you will not need to call the constuctor each time, and do not need to add all the skills in the constructor.

SergeyDev1996 (changes requested):

MasterpieceElbow (approved):

PR #46, opened 2022-05-24
'Solution'

YuraBeseda:


MasterpieceElbow (approved):

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def __init__(self, name):
+        super().__init__(name)
```
MasterpieceElbow, line 44: You don't need to create `__init__` if all it does is trigger parent `__init__` with the same arguments

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #58, opened 2022-05-25
Solution

mariia-movchan:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str):
+        super(FullStackDeveloper, self).__init__(name)
```
MasterpieceElbow, line 43: You don't need to create `__init__` if all it does is call parent `__init__` with the same arguments.

MasterpieceElbow (changes requested):

app/main.py
mariia-movchan: > You don't need to create `__init__` if all it does is call parent `__init__` with the same arguments.

fixed :)

MasterpieceElbow (approved):

PR #62, opened 2022-05-26
Solution

Vasyl-Smutok:


app/main.py
```diff
@@ -1 +1,56 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str):
+        self.name = name
+
+    def learn_skill(self, skill: str):
+        self.__class__.skills.append(skill)
```
IvanRamyk, line 8: Why do you use `__class__` attribute here?

IvanRamyk (changes requested):

app/main.py
Vasyl-Smutok: I thoght it was a class attribute and needed there __class__,  i was wrong(

Vasyl-Smutok: but I understood my mistake)

app/main.py
```diff
@@ -1 +1,56 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
IvanRamyk, line 2: `skills` should be an object attribute

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #69, opened 2022-06-12
Solution

Splend1ed:


app/main.py
```diff
@@ -1 +1,57 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Abnormaltype, line 2: `skills` should be an object attribute

Abnormaltype (approved):
Good job!

PR #70, opened 2022-06-14
Kazakov

artemkazakov947:
I'm not sure about FullStackDeveloper __init__

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -65,4 +65,3 @@ class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
         print(f'{self.name} started creating a web application...')
         super().create_powerful_api()
         super().create_awesome_web_page()
-jk,j
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,67 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name, skills=None):
+        self.name = name
+        self.skills = [] if skills is None else skills
+
+    def learn_skill(self, skill):
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name, skills=None):
+        super().__init__(name, skills=skills)
+        self.skills = [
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,67 @@
...
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS"
+        ]
+
+    def create_web_application(self):
+        print(f'{self.name} started creating a web application...')
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
Abnormaltype, line 67: you don't need call `super()`

app/main.py
```diff
@@ -1 +1,67 @@
...
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS"
+        ]
+
+    def create_web_application(self):
+        print(f'{self.name} started creating a web application...')
+        super().create_powerful_api()
```
Abnormaltype, line 66: the same here

app/main.py
```diff
@@ -1 +1,67 @@
...
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name, skills=None):
+        super().__init__(name, skills=skills)
+        self.skills = [
```
Abnormaltype, line 55: in this class, you should not to create `self.skills`

Abnormaltype (changes requested):

app/main.py
artemkazakov947: I knew it!

Abnormaltype: 😂😂

artemkazakov947: I leaved two version. In last version I deleted __init__ for FullStackDeveloper. And it still works

Abnormaltype (approved):
Good job)

PR #76, opened 2022-06-30
Solution

VladyslavKlevanskyi:


app/main.py
```diff
@@ -1 +1,56 @@
-# write your code here
+class SoftwareEngineer:
+
+    skills = []
```
Abnormaltype, line 3: `skills` should be an instance attribute

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str):
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
Abnormaltype, line 43: it's not correct, use inheritance)

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str):
+        super().__init__(name)
+        self.skills =\
+            BackendDeveloper(name).skills + \
+            FrontendDeveloper(name).skills
```
Abnormaltype, line 45: with inheritance, you don't need to call `__init__`, try to implement this

Abnormaltype (changes requested):

app/main.py
VladyslavKlevanskyi: Sorry, dont't udertand you. What do you mean?

Abnormaltype: you should use `__init__ ` your parents

VladyslavKlevanskyi: I understand your idea, but I can't imagine how to do it )

Abnormaltype: what about method `extends()` ?)

VladyslavKlevanskyi: I still don’t understand how it will help ) For me, the whole problem lies in the fact that `skills `is a method of an instance and not a class. I don’t understand how to access the `skills `of an instance, which, for example, is in the `BackendDeveloper `class, without creating an instance there.

VladyslavKlevanskyi: Are lines 41 and 42 correct? Maybe I'm looking in the wrong place

VladyslavKlevanskyi: Is it ok? 
изображение

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_powerful_api(self):
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str):
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
IvanRamyk, line 33: Add the list of the required skills to the attribute. In that case, you shouldn't create a constructor for `FullStackDeveloper` class
```suggestion
        self.skills += ["Java", "Android studio"]
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #77, opened 2022-07-06
solution

AlexBOVUKR:


app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def learn_skill(self, skill):
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
IvanRamyk, line 16:
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```
Add the list of the required skills to the attribute. In that case, you shouldn't create a constructor for FullStackDeveloper class

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
IvanRamyk, line 49: Could be added by parent classes

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,54 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name):
```
IvanRamyk, line 47: In this class, constructor is redundant because constructors of the `super()` classes are called by default

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #84, opened 2022-07-15
Solution

UaMermaid:


app/main.py
```diff
@@ -1 +1,55 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Abnormaltype, line 2: `skills` - should be an instance attribute

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,16 +1,23 @@
...
+        self.skills = skills
+        if isinstance(self, FrontendDeveloper):
+            self.skills = ["JavaScript", "HTML", "CSS"]
+        if isinstance(self, BackendDeveloper):
+            self.skills = ["Python", "SQL", "Django"]
+        if isinstance(self, AndroidDeveloper):
+            self.skills = ["Java", "Android studio"]
+        if isinstance(self, FullStackDeveloper):
+            self.skills = ["Python", "SQL", "Django",
+                           "JavaScript", "CSS", "HTML"]
```
Abnormaltype, line 14: you should define `self.skills = ["Python", "SQL", "Django"]` in `BackendDeveloper` constructor, not in `SoftwareEngineer`, also other cases

app/main.py
```diff
@@ -1,16 +1,23 @@
 class SoftwareEngineer:
-    skills = []
 
-    def __init__(self, name: str):
+    def __init__(self, name: str, skills=[]):
```
Abnormaltype, line 3: it's not correct

Abnormaltype (changes requested):

IvanRamyk (approved):
Good job!

PR #87, opened 2022-07-18
'Solution'

AlenOl:


app/main.py
```diff
@@ -1 +1,55 @@
...
+    def create_smooth_mobile_app(self):
+        print(f'{self.name} is creating a mobile app...')
+        return 'Ads every three swipes'
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name):
+        super().__init__(name)
+        FullStackDeveloper.skills = self.skills
```
IvanRamyk, line 48: What does this line do?

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+    def create_web_application(self):
+        print(f'{self.name} started creating a web application...')
+        back_developer = BackendDeveloper(self.name)
```
IvanRamyk, line 52: You shouldn't create a new instance of the parent class from the child. You can call its methods directly because of inheritance
```
self.create_powerful_api()
```

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self):
+        print(f'{self.name} is creating a mobile app...')
+        return 'Ads every three swipes'
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name):
+        super().__init__(name)
```
IvanRamyk, line 47: this is a default constructor implementation
you can remove it and nothing will change

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #92, opened 2022-07-31
'Solution'

yaroslav-demchenko:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().__init__(name)
+        self.skills += ["Java", "Android studio", ]
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name):
```
IvanRamyk, line 42: If a constructor is absent in the class declaration super constructor is called instead. 
So, you can remove this constructor

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #100, opened 2022-08-14
Solution

Kev1nXD:


app/main.py
```diff
@@ -1 +1,70 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name):
+        self.name = name
```
DmytroSvirsa, line 5: put skills into init, so when you inherit for FullStackDeveloper you will not duplicate code

DmytroSvirsa (changes requested):

IvanRamyk (approved):
Good job!

PR #101, opened 2022-08-15
'Solution'

Pavilion4ik:


app/main.py
```diff
@@ -1 +1,51 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name):
+        self.name = name
```
DmytroSvirsa, line 5: skills must be in `__init__`, if so and you don't need line $46

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #102, opened 2022-08-15
'Solution'

Diliaver:


app/main.py
```diff
@@ -1 +1,52 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
IvanRamyk, line 2: list of skills should be an attribute of the concrete developer (not an attribute of the class)

IvanRamyk (changes requested):

app/main.py
Diliaver: Firstly, you should create `SoftwareEngineer` class, 
which will be a parent for all other classes. 
Its constructor should take one parameter: `name` of the 
engineer. **Also, it should have `skills` attribute - 
a list of engineer's skills, such as `"Python"` or `"JavaScript"`.
`skills` should be initialised with empty list.**

Isn't it in the description or I missunderstand something?

IvanRamyk: Maybe the description isn't occurate. But it's meant that an engineer (an instance of the class) should have a list of his/her skills

IvanRamyk: We will fix the description to make it clearer

IvanRamyk (changes requested):
please, fix the problem with attributes.
if you have additional questions ask them on slack

IvanRamyk (approved):
Good job!

PR #105, opened 2022-08-20
Solution

viktor-codes:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name):
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
Abnormaltype, line 45: you should remove this `init` here

Abnormaltype (changes requested):

app/main.py
rurakite: hey. could you explain which init do you mean exactly, please?

Abnormaltype: init in the `FullStackDeveloper` class

rurakite: when I remove this

def __init__(self, name):
        super().__init__(name)
I can't pass two tests

Abnormaltype: because you should extend these skills)

rurakite: give me an advice please. I don't understand how to do that(

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str):
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name):
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
Abnormaltype, line 15:
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

rurakite: thank you! it seems working now. please check the update.

Abnormaltype (approved):
Well done!

rurakite:
Could you also check the drone-delivery problem, please.

PR #106, opened 2022-08-23
Solution

Paul-Starodub:


app/main.py
```diff
@@ -1 +1,58 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Abnormaltype, line 2: skills - should be instance attribute

Abnormaltype (changes requested):

app/main.py
Paul-Starodub: Firstly, you should create SoftwareEngineer class, which will be a parent for all other classes. Its constructor should take ONE parameter: name of the engineer.

Abnormaltype: yes, you take only one parameter name, but skills - it's not a parameter, it's an attribute.

Abnormaltype (approved):
Good job!

PR #111, opened 2022-09-08
'Solution'

davidkrivko:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+full_stack_dev = FullStackDeveloper("Tom")
+print(full_stack_dev.skills)
+full_stack_dev.create_web_application()
+
+android_dev = AndroidDeveloper("Beth")
+print(android_dev.skills)
```
DmytroSvirsa, line 49: remove unused code

app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
DmytroSvirsa, line 2: make it as self attribute

DmytroSvirsa (changes requested):

app/main.py
davidkrivko: put it in the constructor?

Abnormaltype: yep

Abnormaltype (changes requested):
put in the init

PR #124, opened 2022-09-16
Solution

sbublinskyi:


mltdll (approved):
Good solution, but you may as well add type annotation. For example: 
```python
def create_smooth_mobile_app(self) -> str:
```
```python
```

app/main.py
```diff
@@ -1 +1,63 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name):
```
Nattalli, line 2: add type annotation for the `name` variable and for function

app/main.py
```diff
@@ -1 +1,63 @@
...
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name):
```
Nattalli, line 11: add type annotation for the `name` variable and for function

app/main.py
```diff
@@ -1 +1,63 @@
...
+        )
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name):
```
Nattalli, line 28: add type annotation for the `name` variable and for function

app/main.py
```diff
@@ -1 +1,63 @@
...
+            ])
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name):
```
Nattalli, line 44: add type annotation for the `name` variable and for function

Nattalli (changes requested):

Nattalli (approved):
Well done! 👍🏻

PR #134, opened 2022-09-20
Solution

StasOkhrym:


app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str,):
```
mltdll, line 3: You have a trailing comma here, remove it.

mltdll (changes requested):

app/main.py
StasOkhrym: Yes, sir!

mltdll (approved):

mltdll (approved):
Wow, my approve was so fast it turned out to be pre-emptive. I'll try not to check first from now on.

Nattalli (approved):
Good job!

PR #135, opened 2022-09-20
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -1 +1,68 @@
...
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str):
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super(FrontendDeveloper, self).__init__(name=name)
```
taras-andruschenko, line 14: there is no need to specify "Classname, self" in super() statement, but it is up to u

taras-andruschenko (approved):

app/main.py
olia-trofymchuk: thank you for your review!

app/main.py
```diff
@@ -1 +1,68 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Nattalli, line 2: skills should be an attribute of the instance not class

Nattalli (changes requested):

Nattalli (approved):
Good job!

PR #139, opened 2022-09-21
Solution

vladimir-byshevets:


app/main.py
```diff
@@ -1 +1,57 @@
-# write your code here
+class SoftwareEngineer:
+
+    skills = list()
```
IvanRamyk, line 3: `skills` should be an attribute of the instance

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    def __init__(self, name):
+        self.name = name
```
IvanRamyk, line 6:
```suggestion
        self.name = name
        self.skills = []
```

app/main.py
```diff
@@ -1 +1,57 @@
...
+    @classmethod
+    def learn_skill(cls, skill):
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name):
+        self.name = name
+        FrontendDeveloper.skills = ["JavaScript", "HTML", "CSS"]
```
IvanRamyk, line 17:
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

IvanRamyk: Add new skills to the attribute of the instance

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #143, opened 2022-09-22
SKantor solution

KantorSerhiy:


app/main.py
```diff
@@ -1 +1,59 @@
...
+            "Android studio",
+        ])
+
+    def create_smooth_mobile_app(self):
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper,):
+    def __init__(self, name: str):
```
IvanRamyk, line 53: if the constructor is absent in class the parent's one is called by default 

so this constructor is extra

IvanRamyk (changes requested):
remove constructor from the last class

app/main.py
KantorSerhiy: understand, thank you!

IvanRamyk (approved):
Well done!

PR #156, opened 2022-09-24
Created class software engin with multiple inheritance

alexzubkoff:


FranchukS (changes requested):

app/main.py
```diff
@@ -1 +1,63 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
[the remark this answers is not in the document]
alexzubkoff: Thanks! Agree with you, i reread the task and
 saw engineer attr with default set.

FranchukS (approved):
Well done!

alexzubkoff:
Thank you, for your attention!

Вс, 25 сент. 2022 г. в 22:33, Franchuk Sergii ***@***.***>:

app/main.py
```diff
@@ -1 +1,63 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(cls, skill: str) -> None:
```
Nattalli, line 7: why do you use `cls` variable? `self` is good enough for this

app/main.py
```diff
@@ -1 +1,63 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Nattalli, line 58: this method is not needed at all

Nattalli (changes requested):

alexzubkoff:
Thanks! Cls instead self, because class method remove and cls forgot to
change.

Пн, 26 сент. 2022 г. в 22:29, Nataliia Zakharchuk ***@***.***

Nattalli (approved):
Awesome! 😎

PR #163, opened 2022-09-25
Solution

LiudmylaKulinchenko:


mltdll (approved):

app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
```
Nattalli, line 9: I am not sure that `@classmethod` is really needed here. You can change `skills` value with `self` variable as well

app/main.py
```diff
@@ -1 +1,53 @@
...
+        "Android studio",
+    ]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    skills = FrontendDeveloper.skills + BackendDeveloper.skills
```
Nattalli, line 48: it is unnecessary as it will be done automatically the same as it's a child class
remove it

Nattalli (changes requested):

Nattalli (approved):
Good job! 😸

PR #168, opened 2022-09-26
'Solution'

OvDima1:


app/main.py
```diff
@@ -1 +1,47 @@
-# write your code here
+class SoftwareEngineer:
+
+    skills = []
```
Vasyl-Poremchuk, line 3: I think that the class attribute is not necessary here, since you created the attribute of the instance of the class (line 7: `self.skills = []`), and in the method `learn_skill` you add values to it

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):
GJ!

Nattalli (approved):
Well done!

PR #169, opened 2022-09-26
Solution

Vadosh:


app/main.py
```diff
@@ -1 +1,45 @@
-# write your code here
+class SoftwareEngineer:
+
```
HStakhiv, line 2: Why is there an empty line?

HStakhiv (approved):
Everything else looks fine to me

PR #171, opened 2022-09-26
Solution

graandda:


app/main.py
```diff
@@ -1 +1,49 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django",
+                       "JavaScript", "HTML", "CSS"]
```
Nattalli, line 44: it is unnecessary as it will be done automatically the same as it's a child class
remove it

Nattalli (changes requested):

app/main.py
graandda: Ok, Thank you)

Nattalli (approved):
Good job! 😸

PR #176, opened 2022-09-26
Solution

pazyn404:


app/main.py
```diff
@@ -1 +1,53 @@
...
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("Python")
+        self.learn_skill("SQL")
+        self.learn_skill("Django")
```
Vadosh, line 27: you can use extend() method once to add three elements to list except calling method three times

app/main.py
```diff
@@ -1 +1,53 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super(FullStackDeveloper, self).__init__(name)
+        super(FullStackDeveloper, self).__init__(name)
+
```
Vadosh, line 49: I could be wrong, but maybe using super().__init__(name) is enough or just don't use __init__ in this class and rely on MRO ;)

Vadosh (approved):

app/main.py
```diff
@@ -1 +1,53 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
+        self.learn_skill("HTML")
+        self.learn_skill("CSS")
```
Nattalli, line 15: look at this checklist:
https://github.com/mate-academy/py-software-engineers/blob/master/checklist.md

Fix these lines as it mentioned at the file

app/main.py
Nattalli: true

app/main.py
```diff
@@ -1 +1,53 @@
...
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("Java")
+        self.learn_skill("Android studio")
```
Nattalli, line 38: and here

app/main.py
Nattalli, line 49: this code is unnecessary as it will be done the same as it's a child class
remove it

Nattalli (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        super().__init__(name)
```
Nattalli, line 43: no, I mean that you can remove `__init__` method at all because you don't do anything here
so as `FullStackDeveloper` is a child class - this methods will be called automatically

Nattalli (changes requested):

Nattalli (approved):
Good job! 😸

PR #188, opened 2022-10-01
Solution

marina-bidenko:


app/main.py
```diff
@@ -1 +1,64 @@
...
+            "Android studio",
+        ])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    AndroidDeveloper,
```
MasterpieceElbow, line 52: A fullstack developer shouldn't be a child of the android developer.

app/main.py
```diff
@@ -1 +1,64 @@
...
+    BackendDeveloper,
+    FrontendDeveloper
+):
+
+    def __init__(self, name: str) -> None:
+        super(FullStackDeveloper, self).__init__(name)
+        self.skills = ['CSS', 'Django', 'HTML', 'JavaScript', 'Python', 'SQL']
```
MasterpieceElbow, line 59: You don't have to assing `self.skills` here. It is already done in the `super()` calls.

MasterpieceElbow (changes requested):

PR #190, opened 2022-10-06
solution

taras-zelinskyi:


app/main.py
```diff
@@ -1 +1,60 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript",
```
Abnormaltype, line 12: it should be an instance attribute

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper,
+                         BackendDeveloper):
+
+    def __init__(self, name: str) -> None:
```
IvanRamyk, line 44: this declaration is extra
if the constructor is absent the parent's constructor is called by default

so, it's better to remove it

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #195, opened 2022-10-09
Solution of py-delivery-drone

YLobunets:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+from __future__ import annotations
+
+if __name__ == "__main__":
+    from developers import SoftwareEngineer, FrontendDeveloper, \
```
IvanRamyk, line 4: please, move classes to main file

IvanRamyk (changes requested):

app/main.py
YLobunets: Everything is transferred to main

app/main.py
```diff
@@ -1 +1,78 @@
...
+            ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+        self.learn_skills(default_skills)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+# Code test block
```
IvanRamyk, line 69: It's better to remove comments

IvanRamyk (approved):
Good job!

PR #212, opened 2022-10-31
Solution

Vanya2389:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
IvanRamyk, line 23: it's better to **add** new skills to attribute but not **set**
using this approach try to avoid duplicating skills in full-stack constructor

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #215, opened 2022-11-04
Solution

mish-sk:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name)
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
IvanRamyk, line 41: an extra declaration
the parent's constructor is called. by default if the child's constructor isn't written

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #216, opened 2022-11-08
Done

DevSpaciX:


app/main.py
```diff
@@ -1 +1,74 @@
...
+                         BackendDeveloper,
+                         SoftwareEngineer):
+    skills = [
+        "Python",
+        "SQL",
+        "Django",
+        "JavaScript",
+        "HTML",
+        "CSS",
+    ]
```
DmytroSvirsa, line 66: make it work thrue inharitanse, make skills self atr

DmytroSvirsa (changes requested):

app/main.py
DevSpaciX: okay )

app/main.py
```diff
@@ -1 +1,64 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
```
DmytroSvirsa, line 5: it also should be self attribute here

app/main.py
```diff
@@ -1 +1,64 @@
...
+class BackendDeveloper(SoftwareEngineer):
+    skills = [
+        "Python",
+        "SQL",
+        "Django",
+    ]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
DmytroSvirsa, line 31: here it is self attribute hand cls attribute

app/main.py
```diff
@@ -1 +1,64 @@
...
+                         SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python",
+                       "SQL",
+                       "Django",
+                       "JavaScript",
+                       "HTML",
+                       "CSS"]
```
DmytroSvirsa, line 59: and after all, make some code changes to not wright this part of the code

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):
GJ

PR #218, opened 2022-11-09
Solution

tinabushuieva:


app/main.py
```diff
@@ -1 +1,59 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    skills = BackendDeveloper.skills + FrontendDeveloper.skills
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
DmytroSvirsa, line 54: an extra declaration
the parent's constructor is called. by default if the child's constructor isn't written

app/main.py
```diff
@@ -1 +1,59 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    skills = BackendDeveloper.skills + FrontendDeveloper.skills
```
DmytroSvirsa, line 51: you can avoid this line if make skills the self attribute

DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
-# write your code here
+class SoftwareEngineer:
+
+    skills = []
```
Serhii-Leonenko, line 3: `skills` should be instance attribute instead of class attribute

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.extend([skill])
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
```
Serhii-Leonenko, line 14: You should return the `__init__` method, and for every skill in skills use method`learn_skill()`. The same in `BackendDeveloper`

app/main.py
Serhii-Leonenko: and remove this line

Serhii-Leonenko (changes requested):

DmytroSvirsa (approved):

PR #239, opened 2022-12-04
Solution

roma-mechanic:


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
+        FrontendDeveloper.create_awesome_web_page(self)
```
OleksandrPozden, line 52: This is a bad practice. Use self to refer to the parent methods

OleksandrPozden (changes requested):

YuriiKindrat (approved):

PR #243, opened 2022-12-06
Solution

Nexxel05:


app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Serhii-Leonenko, line 2: Make `skills` instance's attribute.
And then in child classes append skills to this attribute.

app/main.py
```diff
@@ -1 +1,53 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    skills = BackendDeveloper.skills + FrontendDeveloper.skills
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
```
Serhii-Leonenko, line 49: you can remove this if everything before will be correct

Serhii-Leonenko (changes requested):

OleksandrPozden (approved):

PR #244, opened 2022-12-06
solution

kleotan901:


app/main.py
```diff
@@ -1 +1,50 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+        return self.skills
```
YuriiKindrat, line 9: No need to return self.skills here

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = [
+        "JavaScript",
+        "HTML",
+        "CSS"]
```
YuriiKindrat, line 16: fix the breaking line here, according to the best practices

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = [
+        "Python",
+        "SQL",
+        "Django"]
```
YuriiKindrat, line 27: the same here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #248, opened 2022-12-10
'Solution'

evgenijmartynuk07:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
+
```
OleksandrPozden, line 17:
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]

```
and the same for others

OleksandrPozden (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Serhii-Leonenko, line 45: Redundant init

Serhii-Leonenko (changes requested):

app/main.py
evgenijmartynuk07: def __init__(self, name: str) -> None:
        self.name = name
        
        like this?

OleksandrPozden (approved):

PR #254, opened 2022-12-16
Solution

Isles22:


app/main.py
```diff
@@ -1 +1,68 @@
-# write your code here
+from __future__ import annotations
+
+
+class SoftwareEngineer:
+
+    skills = []
+
```
OleksandrPozden, line 7: skills should  be in the constructor

app/main.py
```diff
@@ -1 +1,68 @@
...
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
```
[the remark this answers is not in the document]
Isles22: thanks

app/main.py
```diff
@@ -1 +1,64 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = BackendDeveloper(self.name).skills + \
+            FrontendDeveloper(self.name).skills
```
YuriiKindrat, line 59: No need to write like this, the skills will be inherited from Back/Front developers classes

YuriiKindrat (changes requested):

app/main.py
Isles22: I'm uderstood, thank you

app/main.py
```diff
@@ -1 +1,62 @@
...
+        ]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
YuriiKindrat, line 57: def __init__ also will be inherited, so there is no need to write it

YuriiKindrat (changes requested):

app/main.py
Isles22: Realy. OK

YuriiKindrat (approved):

PR #264, opened 2022-12-29
Solution

Tania-Kharchuk:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
IvanRamyk, line 42: Super constructor is used by default so no need to create it inside child class

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #282, opened 2023-01-15
'Solution'

StinGER-ua:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += ["JavaScript", "HTML", "CSS"]
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        page = "<h1>Hello world</h1>"
```
katevasilenko, line 17: As you never use this variable but in return only, just return put your string directly to `return`

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += ["Python", "SQL", "Django"]
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        api = "http://127.0.0.1:8000"
```
katevasilenko, line 28: As you never use this variable but in return only, just return put your string directly to `return`

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        app = "Ads every three swipes"
```
katevasilenko, line 39: As you never use this variable but in return only, just return put your string directly to `return`

katevasilenko (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #284, opened 2023-01-16
Solution

PodorogaNatalia:


app/main.py
```diff
@@ -1 +1,42 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Abnormaltype, line 2: it should be an instance attribute

Abnormaltype (changes requested):

ArsenPidhoretskyi (approved):
Well done

PR #286, opened 2023-01-20
'Solution'

Vitalii2906:


app/main.py
```diff
@@ -1 +1,72 @@
-# write your code here
+class SoftwareEngineer:
+    skills = list()
```
ArsenPidhoretskyi, line 2: It must be instance attribute, not class

app/main.py
```diff
@@ -1 +1,72 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
```
ArsenPidhoretskyi, line 8: just method. You do not need to affect all other objects
```
person1 = SoftwareEngineer()
person2 = SoftwareEngineer() 
person1.learn_skill('Python')
```

app/main.py
```diff
@@ -1 +1,72 @@
...
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = [
+        "Python",
+        "SQL",
+        "Django",
+    ]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
ArsenPidhoretskyi, line 35: You can also do self.skills += list of another skills
Change in all places

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #288, opened 2023-01-22
solution

OleksandrGrk:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        if self.__class__ == FullStackDeveloper:
+            self.skills += ["JavaScript", "HTML", "CSS"]
+        else:
+            self.skills = ["JavaScript", "HTML", "CSS"]
```
olena-hul, line 17: It is not needed. Just extend the `self.skills` list from parent class in 1 line

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
olena-hul, line 27: Extend the list, don't reassign it

olena-hul (changes requested):

app/main.py
OleksandrGrk: Actually, I don`t see how...

OleksandrGrk: Done

app/main.py
OleksandrGrk: Done

Vasyl-Smutok (approved):
Well done 👍

PR #290, opened 2023-01-24
Solution

xtina999:


app/main.py
```diff
@@ -1 +1,48 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str, ) -> None:
```
ArsenPidhoretskyi, line 3:
```suggestion
    def __init__(self, name: str) -> None:
```

ArsenPidhoretskyi (approved):
Good! Also, no need empty line after class declaration and `__init__`

PR #296, opened 2023-01-28
Solution

Oleksy-Syrota:


app/main.py
```diff
@@ -1 +1,53 @@
...
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+# new_skills = ["one", "two"]
+# engieenier = BackendDeveloper("Max")
+# print(engieenier.skills)
+# engieenier.learn_skill(new_skills)
+# print(engieenier.skills)
```
YuriiKindrat, line 53: remove unnecessary comments here

app/main.py
```diff
@@ -1 +1,53 @@
...
+    skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
YuriiKindrat, line 41: no need to rewrite skills here, make it so skill will inherit from inherited classes

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
-# write your code here
+class SoftwareEngineer:
+
+    skills = []
```
vsmutok: skills variable should be an attribute of the instance not class.

DaTrEvTeR (approved):
Well done!

PR #298, opened 2023-01-29
'Solution'

contritum:


app/main.py
```diff
@@ -1 +1,58 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.extend(skill.split())
```
ArsenPidhoretskyi, line 8: engineer = SoftwareEngineer()
engineer.learn_skill("Django Rest Framework")

What do you expect as a result?
As for me, skill == ["Django Rest Framework"] but your code will say another

app/main.py
```diff
@@ -1 +1,58 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
ArsenPidhoretskyi, line 2: Why class attribute due to task it just instance attribute

app/main.py
```diff
@@ -1 +1,58 @@
...
+    FrontendDeveloper
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
+        FrontendDeveloper.create_awesome_web_page(self)
```
ArsenPidhoretskyi, line 53: Use .self why. Nobody not send self as parameter if it already have this method

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    AndroidDeveloper,
+    BackendDeveloper,
+    FrontendDeveloper
```
ArsenPidhoretskyi, line 44: Please explain why you need inherit from AndroidDeveloper?

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
ArsenPidhoretskyi, line 34: Use += instead

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
ArsenPidhoretskyi, line 24: Use += instead

app/main.py
```diff
@@ -1 +1,58 @@
...
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.extend(skill.split())
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
ArsenPidhoretskyi, line 14: Use += instead

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well DOne!

PR #300, opened 2023-01-29
Solution

romkapomka12:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
ArsenPidhoretskyi, line 33: In all places is += but in this no. Why?
If someone wants create frontend + android?

ArsenPidhoretskyi (changes requested):

app/main.py
romkapomka12: did not pay attention to this. Then FullStackDeveloper should be a child class AndroidDeveloper?

andrewbelo (approved):

PR #301, opened 2023-01-30
software engigeers

VKungin:


app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Abnormaltype, line 2: should be the attribute of the instance

Abnormaltype (changes requested):

katevasilenko (approved):
Look good to me

PR #307, opened 2023-02-02
Solution

oleksandr-lysanets:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
ArsenPidhoretskyi, line 13: Use += because in child classes you need to use already given skills by previous class
For example FullStack must-have skills from Frontend and Backend

ArsenPidhoretskyi: The same in Backend and Android classes

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done

PR #314, opened 2023-02-08
Solution

manjustice:


bihunva (approved):
Good job!

bihunva (approved):

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 45: You don't really need to override __init__ if it's just a call to super() with the same arguments

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += ["Python", "SQL", "Django"]
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+
```
danylott, line 28: Do you always add an empty line before `return`?
P.S: It's ok if you always do that, but you should stick to this rule (in this task you stick)

danylott: Of course except situations, where the only 1 line in function is starting with `return` :)

danylott (changes requested):

app/main.py
manjustice: Yeah, I always add an empty line before return, maybe sometimes I forget do this. So is it better to do this, or without an empty line?

danylott: Depending on the situation, I'd said. But if you like to do so - just do it everywhere. It's not a critical code-style issue :)

danylott (approved):
LGTM

PR #315, opened 2023-02-08
Solution

MaybeTI:


app/main.py
```diff
@@ -1 +1,58 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["JavaScript",
+                       "HTML",
+                       "CSS"]
```
manjustice, line 15: Here you must use `self.skills +=`

app/main.py
```diff
@@ -1 +1,58 @@
...
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["Python",
+                       "SQL",
+                       "Django"]
```
manjustice, line 27: same

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["Java",
+                       "Android studio"]
```
manjustice, line 38: same

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["Python",
+                       "SQL",
+                       "Django",
+                       "JavaScript",
+                       "HTML",
+                       "CSS"]
```
manjustice, line 53: So these rows are unnecessary.

manjustice (changes requested):

manjustice (approved):

app/main.py
danylott: @manjustice as for me - too direct suggestion - you should provide hint why so, or even less that it :)

app/main.py
```diff
@@ -1 +1,52 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills += ["JavaScript",
+                        "HTML",
+                        "CSS"]
```
danylott, line 15: As for me, this style for me is not ideal for separating values into the new line.
Here you can use 1 line `["JavaScript", "HTML", "CSS"]`.
Or separate like `black` suggest (but only, if more values than here):
```python
skills = [
    "First skill",
    "Second skill",
    "Third skill",
    "Fourth skill",
    "Fifth skill",
]
```

app/main.py
```diff
@@ -1 +1,52 @@
...
+                        "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
danylott, line 47: You don't really need this __init__ if it's just a super() call with the same arguments

danylott (changes requested):

YuriiKindrat (approved):

PR #316, opened 2023-02-08
Solution

bihunva:


AndereLion (approved):
Good one!

danylott (commented):
LGTM

app/main.py
```diff
@@ -1 +1,48 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
```
danylott, line 2: Sorry, forgot to mention earlier - you should also add here empty line before __init__, or remove it in all other classes - stick to the same code style everywhere

danylott (changes requested):

YuriiKindrat (approved):

PR #319, opened 2023-02-09
Solution

DavidGlusch:


app/main.py
```diff
@@ -1 +1,61 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
NEkropo1, line 2: checklist.md
2. `skills` variable should be an attribute of the instance not class.

app/main.py
```diff
@@ -1 +1,61 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
NEkropo1, line 12: same as above

app/main.py
```diff
@@ -1 +1,61 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = ["Python", "SQL", "Django"]
```
NEkropo1, line 23: same

app/main.py
```diff
@@ -1 +1,61 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    skills = ["Java", "Android studio"]
```
NEkropo1, line 34: Same )

app/main.py
```diff
@@ -1 +1,61 @@
...
+        super().__init__(name)
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    skills = [
```
NEkropo1, line 46: and here

NEkropo1 (changes requested):

NEkropo1 (approved):
Gj!

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 47: You don't really need to override __init__ if it's just a call to super() with the same arguments

danylott (changes requested):

app/main.py
NEkropo1: explicit over implicit? I thought it's a good code case, if we want some class to BE initiated, to have init method by default, even if it only calls super with same args?

danylott: Oh, btw, see a great name for a commit :)

danylott (approved):
LGTM

app/main.py
NEkropo1: It's just one of base cases, but what if we get throughout your favorite 10000 lines of code, and we see some abstract methods and inherited, and other things like those. It would take much more time refactoring these, when u can just see 
```
#bad
class A(ABC):
    pass
    
class B(A):
    pass
    
class D:
    def method()
    
class C(D):
    pass
# instead of
#good
class A(ABC):
    pass
    
class B(A):
    pass
    
class D:
    def method()
    
class C(D):
    def __init__:
    super
```

So I'm pretty sure, `Readability`, `Explicit`, `Beautiful` and even `Simple`, 'cause making refactoring simple is one of cases, gonna be much better than your idea. I will refactor it according to your vision, but your arguments weren't good enough to prove me as an aksioma, sry (

NEkropo1: anyway YAWNI in exact case

danylott: I'm suggesting my thoughts, of course, you can stick to some of your rules, and even more other mentors can have their own opinion, so it's fully fine :)

PR #321, opened 2023-02-09
Solution

NEkropo1:


app/main.py
```diff
@@ -1 +1,65 @@
...
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        back_skills = [
+            "Python",
+            "SQL",
+            "Django",
+        ]
+        for skill in back_skills:
```
Maytreyaya, line 34: same here

app/main.py
```diff
@@ -1 +1,65 @@
...
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.name = name
+        android_skills = [
+            "Java",
+            "Android studio",
+        ]
+        for skill in android_skills:
```
Maytreyaya, line 50: same

app/main.py
```diff
@@ -1 +1,65 @@
...
+            self.learn_skill(skill)
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
Maytreyaya, line 60: what are skills of full stack dev?

Maytreyaya (changes requested):

app/main.py
```diff
@@ -1 +1,65 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
```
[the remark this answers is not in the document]
NEkropo1: it's not about proper function, it's about concatenating lists for keeping code dry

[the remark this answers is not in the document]
NEkropo1: read checklist ) 
>
1. Use the `extend()` method or `+=` operator to add **multiple** values to a single list.

app/main.py
```diff
@@ -1 +1,65 @@
...
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        front_skills = [
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
+        for skill in front_skills:
```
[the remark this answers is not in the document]
NEkropo1: agree, but if we do that, our learn_skill method won't be any useful in code, idk.

app/main.py
NEkropo1: learn multiple inheritance )

NEkropo1: learn inheritance ) 
image

DmytryTong (approved):
Nice job) In this particular case you could do inits without local variables, but it's seems that you know, what you doing in generel so - it's ok)

NEkropo1:
> Nice job) In this particular case you could do inits without local variables, but it's seems that you know, what you doing in generel so - it's ok)
Overall idea was in clearing class memory and object memory when u create object, while having local access to list of variables for further changes...

app/main.py
```diff
@@ -1 +1,62 @@
...
+        self.skills.extend(android_skills)
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
danylott, line 57: You don't really need this __init__ if it's just a super() call with the same arguments

app/main.py
danylott: @NEkropo1, please, be a bit more kind, when answering questions :)

app/main.py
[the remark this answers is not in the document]
danylott: Agree with @NEkropo1, here append is ok, because only 1 value

danylott (changes requested):

app/main.py
NEkropo1: Still same question as here:
https://github.com/mate-academy/py-software-engineers/pull/319/files/be88e93d9e570b773a34b3936b490aa33469c698

app/main.py
NEkropo1: Didn't try to insult, it was just a simple suggestion in my opinion and a screenshot of result... Mb need a little bit more understanding of politeness  (

app/main.py
```diff
@@ -1 +1,59 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.name = name
```
YuriiKindrat, line 43: no need to rewrite name attribute here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #322, opened 2023-02-10
Solution

ZabFTFT:


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
Zenionoi, line 47: The  `self.skills` attribute can be inherited from parent classes

Zenionoi (approved):

app/main.py
danylott: @Zenionoi if you have such a big suggestion - I'd like you to `reject` the solution, because your comment is big enough to reject :)

app/main.py
```diff
@@ -1 +1,56 @@
...
+        self.learn_skill("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 51: You don't really need this __init__ here, if it's only the super() call with the same arguments

danylott (changes requested):

app/main.py
ZabFTFT: Wow, okay, I didn't figure out! Thank you!

AnnaBoichenko (approved):

app/main.py
```diff
@@ -1 +1,54 @@
...
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
+        self.learn_skill("HTML")
+        self.learn_skill("CSS")
```
YuriiKindrat, line 17: make ur code DRY use += or extend method here

app/main.py
```diff
@@ -1 +1,54 @@
...
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("Python")
+        self.learn_skill("SQL")
+        self.learn_skill("Django")
```
YuriiKindrat, line 30: the same

app/main.py
```diff
@@ -1 +1,54 @@
...
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("Java")
+        self.learn_skill("Android studio")
```
YuriiKindrat, line 42: same

YuriiKindrat (changes requested):

PR #323, opened 2023-02-10
Solution

Zenionoi:


kuznetsb (approved):
++++

app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = list()
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.append(skill)
```
danylott, line 9: Interesting idea - that skills should have only unique skills, that wasn't mentioned in task description, but if it was - it would be great to use `set` instead of `list` here - and then no additional check is required :)

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 48: You don't really need this __init__ if it's just a super() call with the same arguments

app/main.py
danylott: But no need to fix it here, because it doesn't mentioned in task description

PR #326, opened 2023-02-10
Software Engineers Solution

Maytreyaya:


app/main.py
```diff
@@ -1 +1,58 @@
...
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper,
+                         BackendDeveloper,
+                         SoftwareEngineer):
```
Morty67, line 49: as i know, we need use only  FrontendDeveloper and BackendDeveloper for FullStack
 SoftwareEngineeri t's already default for FrontendDeveloper and BackendDeveloper

Morty67 (changes requested):

Morty67 (approved):

app/main.py
```diff
@@ -1 +1,57 @@
...
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper,
+                         BackendDeveloper):
```
danylott, line 48: Can be written in 1 line

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "CSS", "HTML"]
```
danylott, line 52: You can notice here, that you duplicate some skills here, and in FrontendDeveloper & BackendDeveloper.
Try to think, how you can reuse those skills here, so if you inherit from some developer - all skills he/she has will be added to your skills

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.skills = []
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.extend([skill])
```
YuriiKindrat, line 11: u can use append for adding only one skill here

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #327, opened 2023-02-10
Solution

SergiiMachulin:


app/main.py
```diff
@@ -1 +1,50 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
```
WDemidenko, line 7: Rewrite this part, according to Checklist

WDemidenko (changes requested):

app/main.py
SergiiMachulin: Not agree, according to README.md SoftwareEngineer class should have one method: learn_skill that takes a skill of string type and should add it to the skills list. And checklist example of bad practice is about duplicating code and adding miltiple values, not about using append or expand methods of list. Below condition is from checklist you mentioned: 
```Use the extend() method or += operator to add multiple values to a single list```

Comment (WDemidenko) [reply app/main.py:7]: My bad

**Review by WDemidenko — APPROVED** [review]

File: app/main.py (hunk 1 above)
Comment (SergiiMachulin) [reply app/main.py:7]: no problem

File: app/main.py [hunk 2]
```diff
@@ -1 +1,50 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    FrontendDeveloper,
+    BackendDeveloper,
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 45: You don't really need this __init__ if it's just a super() call with the same arguments

danylott (changes requested):

danylott (approved):
LGTM

PR #328, opened 2023-02-10
Solution

WDemidenko:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
ramsty, line 42: This method is unnecessary here, you'll inherit the name from parent class.

ramsty (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.extend([skill])
```
danylott, line 8: An interesting decision, it's really logical, that the skills list shouldn't contain duplicates, but that of course hasn't been written in the task description.
But if it was, the great solution would be not to check for existence each time, but to simply use `set` instead of `list`

danylott (approved):

PR #330, opened 2023-02-11
'Solution'

alexlega:


app/main.py
```diff
@@ -43,10 +43,10 @@ def create_smooth_mobile_app(self) -> str:
 
 class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
     def __init__(self, name: str) -> None:
-        super(FullStackDeveloper, self).__init__(name)
+        super().__init__(name)
         self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
usrofgh, line 47: your code works correctly and pass all tests. But i'm not sure about logic of this code.
in FullStackDeveloper you just manually wrote skills from parents classes. Probably the main idea of this task to add in child classes those skills which there are in parents classes. Also in checklist there is hint about it

usrofgh (approved):

app/main.py
alexlega: yeah, my bad, I forgot to check the checklist. And I thought about code duplicate in my case, but hadn't a good idea how I could solve it

app/main.py
```diff
@@ -1 +1,61 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self,
+                 name: str
+                 ) -> None:
```
danylott, line 5: It's my personal taste, but this kind of code-style for functional parameters are not so great.
I suggest 2 other options:
1st:
Simply in 1 line, because we have only 2 parameters here (no need to make a new lines):
`def __init__(self, name: str) -> None:`
2nd:
if a lot of arguments - you can use `black` codestyle:
```python
def __init__(
    self,
    param1,
    param2,
    param3,
    param4,
):
    pass
```

app/main.py
```diff
@@ -1 +1,61 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self,
+                 name: str
+                 ) -> None:
+        super().__init__(name)
```
danylott, line 56: You don't really need to override __init__ if it's just a call to super() with the same arguments

app/main.py
alexlega: okay, thanks for the advice!) Is second option usable only for __init__ or even methods and functions?

NazarHladaniuk (approved):

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super(AndroidDeveloper, self).__init__(name)
```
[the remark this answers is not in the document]
YuriiKindrat: no need to write class name + self in super method

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #331, opened 2023-02-11
Solution DonchykD

DmytryTong:


yura-hudzovskyi (approved):
it`s so good

app/main.py
```diff
@@ -1 +1,66 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.extend([skill])
```
danylott, line 8: if you're adding only 1 element, there is a better `append` method

app/main.py
```diff
@@ -1 +1,66 @@
...
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
+        self.skills.extend([
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ])
```
danylott, line 20: It's not required, but you can write this in 1 line, it won't be really long here (not required to fix, if you disagree)

app/main.py
```diff
@@ -1 +1,66 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 61: You don't really need to override __init__ if it's just a call to super() with same arguments

danylott (changes requested):

app/main.py
DmytryTong: In a checklist there was exactly an extend, but - ok) Fixed.

app/main.py
DmytryTong: I'll leave it like it was, cause if there is a longer lists - code would look different. It's could make it not easy to read at some point.

app/main.py
DmytryTong: Fixed.

app/main.py
```diff
@@ -1 +1,63 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
```
YuriiKindrat, line 15: no need to rewrite self.name attribute here, u are using super for this here

app/main.py
```diff
@@ -1 +1,63 @@
...
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
```
YuriiKindrat, line 31: the same is here

app/main.py
```diff
@@ -1 +1,63 @@
...
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
```
YuriiKindrat, line 47: same

YuriiKindrat (changes requested):

app/main.py
DmytryTong: Fixed.

app/main.py
DmytryTong: Fixed.

PR #334, opened 2023-02-12
solution

usrofgh:


inrock1 (approved):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 42: You don't really need this __init__ if it's just a super() call with the same arguments

danylott (changes requested):

app/main.py
usrofgh: yes i've got it yet)
Just in [one task](https://mate.academy/learn/python-core/python-core-oop-multiple-inheritance#/practice/python_bike_shop) on mate academy the platform didn't accept the solution cause it [required ](https://prnt.sc/xOj3HPLu6HP5) ```__Init__```
Thus I've wrote the code like it. And it's so I asked about init in child classes on the lection

YuriiKindrat (approved):

PR #335, opened 2023-02-12
Solution

yura-hudzovskyi:


oksana0410 (approved):

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
danylott, line 43: You should try to reuse the skills you've provided in BackendDeveloper & FrontendDeveloper class. And not repeat them here. Think how to do the logic, that if we inherit from some developer - we're getting those skills this specific type of developer has

danylott (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
YuriiKindrat, line 42: no need to initialize init method here

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
+        FrontendDeveloper.create_awesome_web_page(self)
```
YuriiKindrat, line 47: use self instead of class name

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
YuriiKindrat, line 33: what if we will use this class as Front/Back dev. calsses? add extend method here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #337, opened 2023-02-12
solution

amber-marichi:


oksana0410 (approved):

app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = list()
```
danylott, line 5: It's better to use `[]` instead of `list()` - just my opinion, but for me `[]` is much more readable

danylott (approved):

PR #338, opened 2023-02-12
'Solution'

inrock1:


oigod (approved):
GT

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 42: You don't really need to override __init__ if it's just a call to super() with the same arguments

danylott (changes requested):

app/main.py
inrock1: removed unnecessary init

YuriiKindrat (approved):

PR #340, opened 2023-02-13
solution

Yulia-Mala:


app/main.py
```diff
@@ -1 +1,52 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
```
ZhAlexR, line 5: Are you sure the "skills" should be an instance attribute? 
I mean, maybe, it should be stored like a class variable? For example:
class A:
   some_list = []

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
```
ZhAlexR, line 13: If my suggestion above is true, you will not need call "supper" every time.

ZhAlexR (approved):

app/main.py
danylott: Not really - stick to `checklist.md` please. Instance attribute is correct here

app/main.py
danylott: But then all developers will have same skills :(

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
danylott, line 47: You don't really need this __init__ if it's just a super() call with the same arguments

danylott (changes requested):

ProdyRodion (approved):
awesome! ^-^

YuriiKindrat (approved):

PR #342, opened 2023-02-13
Solution

oigod:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
NEkropo1, line 45: You don't really need to override `__init__` if it's just a call to super() with the same arguments

NEkropo1 (changes requested):

app/main.py
oigod: agree

NEkropo1 (changes requested):

YuriiKindrat (approved):

PR #345, opened 2023-02-13
Solution

VladyslavCherkashyn:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
Danil228, line 43: you need to make that the FullStack skills are taken from the FrontEnd and from the BackEnd

Danil228 (changes requested):

Danil228 (approved):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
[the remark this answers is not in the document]
YuriiKindrat: no need to initialize init method without changes

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
YuriiKindrat, line 47: no need to use super here, u can use self instead

YuriiKindrat (changes requested):

Nattalli (approved):

PR #351, opened 2023-02-13
Added point about overriding __init__ with no additional logic

AlessaQ:


app/main.py
```diff
@@ -1 +1,46 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    # @classmethod
```
me1nyk, line 7: i can see its just a comment, but i am not sure its necessary to use it here

me1nyk (approved):
All good, but please check my comment)

app/main.py
YuriiKindrat, line 7: remove unnecessary comment here

YuriiKindrat (changes requested):

app/main.py
AlessaQ: fixed

AlessaQ: fixed

YuriiKindrat (approved):

PR #352, opened 2023-02-13
Solution

Anatolii-Poznyak:


ostapT (approved):

app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills += skill.split(", ")
```
dogmoony, line 7: for what you are using function "split"?

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript, HTML, CSS")
```
dogmoony, line 13: you can make it easier without calling a method "learn_skill", but use for example ".+= [...]"

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("Python, SQL, Django")
```
dogmoony, line 23: same here

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("Java, Android studio")
```
dogmoony, line 33: same here

dogmoony (changes requested):

app/main.py
Anatolii-Poznyak: So what for we create a method learn_skill in this task, if we never use it?

app/main.py
Anatolii-Poznyak: same here

app/main.py
Anatolii-Poznyak: same here

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #353, opened 2023-02-13
'Solution'

RuslanLopatin:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        api_address = self.create_powerful_api()
+        webpage_code = self.create_awesome_web_page()
+        return f"Web app with API: {api_address}, " \
+               f"Web page code: {webpage_code}"
```
YuriiKindrat, line 47: no need to use return here, this method should just call parents methods

YuriiKindrat (changes requested):

Nattalli (approved):

PR #360, opened 2023-02-13
'Solution'

ykopatko:


app/main.py
```diff
@@ -1 +1,48 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+
```
Yulia-Mala, line 44: you hard code skills for FullStackDeveloper. But what if BackendDeveloper and FrontendDeveloper will have another skills? SoftwareEngineer include learn_skill method  so child classes  instances could have more skills. you need to extend list of skills in BackendDeveloper and FrontendDeveloper classes. Pytests for this tack includes such a cases. Did you have all tests run successfully   ?

Yulia-Mala (changes requested):

app/main.py
ykopatko: Yes, that worked. You may check it by yourself in pycharm. Nevertheless, I have updated the code according to your comment.

> you hard code skills for FullStackDeveloper. But what if BackendDeveloper and FrontendDeveloper will have another skills? SoftwareEngineer include learn_skill method so child classes instances could have more skills. you need to extend list of skills in BackendDeveloper and FrontendDeveloper classes. Pytests for this tack includes such a cases. Did you have all tests run successfully ?

myroslavatrschk (approved):

wQuelS (approved):
GJ

PR #363, opened 2023-02-13
Solution

myroslavatrschk:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
```
inrock1, line 43: it is not needed to add init block in Fullstack class

inrock1 (changes requested):

app/main.py
myroslavatrschk: Oh, thanks! I forget to delete it.

YuriiKindrat (approved):

PR #364, opened 2023-02-13
Solution

dogmoony:


oksanaaam (approved):
good job

app/main.py
```diff
@@ -1 +1,56 @@
...
+    SoftwareEngineer
+):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
YuriiKindrat, line 56: no need to use super here, u can use self instead

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #392, opened 2023-03-13
Solution

V-Shkrobatskyi:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for i in ["JavaScript", "HTML", "CSS"]:
+            self.learn_skill(i)
```
ArsenPidhoretskyi, line 17: You can use extend or += no need to iterate here

ArsenPidhoretskyi: The same in other places

ArsenPidhoretskyi (changes requested):

katevasilenko (approved):
Looks good to me

PR #394, opened 2023-03-18
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
```
vasylhnatiuk, line 46: use `self` instead of `BackendDeveloper` or `FrontendDeveloper`

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
vasylhnatiuk, line 42: it would be better to use super().__init__(name=name)

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
ArsenPidhoretskyi, line 42: This code is not needed

ArsenPidhoretskyi (approved):
Well done!

PR #405, opened 2023-03-30
'Solution'

kucherenkodmytro:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
vasylhnatiuk, line 12: add here and all of the code
```suggestion
        super().__init__(name=name)
```

vasylhnatiuk (changes requested):
Fix it

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills += ["JavaScript", "HTML", "CSS"]
```
katevasilenko, line 13: It’s better to use `.extend()` instead of `+=`

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
katevasilenko, line 42: You can remove this here. Inheritance already did this authomatically

katevasilenko (approved):
Looks good to me, but it would be great if you add additional changes

PR #412, opened 2023-04-06
solution software engineers

vborovets0:


app/main.py
```diff
@@ -1 +1,42 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
olena-hul, line 2: This should be instance attribute

app/main.py
```diff
@@ -1 +1,42 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
olena-hul, line 13: Define those skills in `__init__` and use `extend` method to modify the list, not to override it

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = ["Python", "SQL", "Django"]
```
olena-hul, line 21: The same as for FrontendDeveloper

app/main.py
```diff
@@ -1 +1,42 @@
...
+class AndroidDeveloper(SoftwareEngineer):
+    skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    skills = BackendDeveloper.skills + FrontendDeveloper.skills
```
olena-hul, line 37: After modifying the parent classes as I suggested, you wouldn't need this line

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #420, opened 2023-04-15
solution

alim2709:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
ArsenPidhoretskyi, line 14: Use += or extend
Check readme and checklist
In this case init inside FullStackDeveloper will be redundant because we will modify skills by two child classes

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #427, opened 2023-04-26
'Solution'

PetroHavrysh:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

vasylhnatiuk: You must add them because if a class inherits from two classes, all skills must be added automatically. Fix here and all of the code

app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().__init__(name=name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 41: and delete this method because it is duplicate code from parent classes.

vasylhnatiuk (changes requested):

app/main.py
PetroHavrysh: проте без нього валиться super().__init__(name=name)

app/main.py
PetroHavrysh: проте без нього валиться super().__init__(name=name)

Grayder0152: you can remove 40-42 lines

Grayder0152 (approved):

PR #451, opened 2023-05-16
Develop

IvanKorshunovE:


app/main.py
```diff
@@ -1 +1,57 @@
...
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    skills: list[str] = ["JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills: list[str] = FrontendDeveloper.skills
```
```suggestion
        self.skills: list[str] += ["JavaScript", "HTML", "CSS"]
```

vasylhnatiuk: It is a better variant, and you must add += to add skills if the class will inherit from two or more classes.

vasylhnatiuk: change here and all of the code

app/main.py
```diff
@@ -1 +1,57 @@
...
+        super().__init__(name=name)
+        self.skills: list[str] = AndroidDeveloper.skills
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 50: delete init here, if you use `self.skills: list[str] +=` - it will add all of the skills from BackendDeveloper, FrontendDeveloper

vasylhnatiuk (changes requested):

IvanKorshunovE:
Done

IvanKorshunovE:
Please note that I am not able to write this line exactly like this:

self.skills: list[str] += ["JavaScript", "HTML", "CSS"]

I can't add annotations in this case

IvanKorshunovE:
I can't clearly understand the logic of inheriting attributes. What is happening when I do this?
Screenshot 2023-05-16 at 17 53 51

f1 = FullStackDeveloper("")
print(f1.skills)

As I understand, initially, the __init__ of BackendDeveloper called and 
[] += ["Python", "SQL", "Django"]
 is executed
Then the __init__ of FrontendDeveloper is called and 
["Python", "SQL", "Django"] += ["JavaScript", "HTML", "CSS"]
Resulting in ['Python', 'SQL', 'Django', 'JavaScript', 'HTML', 'CSS']

Do I understand this correctly? Thanks

vasylhnatiuk (commented):
Well done

vasylhnatiuk (approved):

app/main.py
```diff
@@ -1 +1,48 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
```
vasylhnatiuk, line 47:
```suggestion
        self.create_powerful_api()
```

app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().create_awesome_web_page()
```
vasylhnatiuk, line 48:
```suggestion
        self.create_awesome_web_page()
```

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: better to use self

ArsenPidhoretskyi (approved):
Well done!

PR #455, opened 2023-05-20
solution

z1k01ff:


app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
ArsenPidhoretskyi, line 2: skils is instance attribute

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #464, opened 2023-05-27
implemeted solution

Oknametra:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
vasylhnatiuk, line 13: Use += it will change list but not create new. Read readme.md and checklist also, change here and all of the code
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

vasylhnatiuk: It will help remove redundant code inside FullStackDeveloper

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
ArsenPidhoretskyi, line 42: This part of code is not needed. Because it works the same without this code

ArsenPidhoretskyi (changes requested):

app/main.py
Oknametra: hm, never noticed that, thanks!

vasylhnatiuk (approved):

PR #470, opened 2023-05-31
Multiple inheritance sw_engineers

FadieievaA:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        [self.learn_skill(skill) for skill in ["JavaScript", "HTML", "CSS"]]
```
YukihiroSM, line 16: what about using `extend`?

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        [self.learn_skill(skill) for skill in ["Python", "SQL", "Django"]]
```
YukihiroSM, line 26: what about using `extend`?

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        [self.learn_skill(skill) for skill in ["Java", "Android studio"]]
```
YukihiroSM, line 36: what about using `extend`?

YukihiroSM (changes requested):

app/main.py
FadieievaA: I thought we suppose to use class instance methods. but, yeah, I can change that

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Bohdan-Salamakha, line 45: u don't need __init__ method here

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #472, opened 2023-06-01
Solution (would be glad if you could tell me how to append more than …

volodymyr-komarnyckyi:
…one element at a time (lines: 14-16 26-28 38-39)). I tried to use *args but nothing good came of it

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,50 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.append(skill)
+
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
```
LiudmylaKulinchenko, line 14: Here u can do like this `self.skills += ["JavaScript", "HTML", "CSS"]` or use `extend` to write less code

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥😎

PR #476, opened 2023-06-03
Solution

Tarasidze:
:)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,55 @@
-# write your code here
+from __future__ import annotations
+
+
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = FrontendDeveloper.skills
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,55 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
YuriiKindrat, line 14: no need to create class attr here) u can make this inside skills
also try to use += for list

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
```
YuriiKindrat, line 44: same here

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = BackendDeveloper.skills + FrontendDeveloper.skills
```
YuriiKindrat, line 50: if u use += for skill there is won't be need to write init method here

YuriiKindrat (changes requested):

app/main.py
Tarasidze: every time I learn something new)

PR #491, opened 2023-06-05
Solution

Gvilion:


app/main.py
```diff
@@ -1 +1,48 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str, *args) -> None:
+        self.name = name
+        self.skills = [*args]
```
LiudmylaKulinchenko, line 4: remove args, this can lead to unexpected behavior and type annotations are missed

LiudmylaKulinchenko (changes requested):

PR #501, opened 2023-06-07
Solution

OleksiiKiselov1996:


app/main.py
```diff
@@ -1 +1,42 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
```
vsmutok, line 9:
```suggestion
    def __init__(self, name: str) -> None:
        self.name = name
        self.skills = []

    def learn_skill(self, skill: str) -> None:
        if skill not in self.skills:
            self.skills.append(skill)
```

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
```suggestion
    def __init__(self, name: str) -> None:
        super().__init__(name)
        self.skills += ["JavaScript", "HTML", "CSS"]
```

vsmutok: please do this wherever necessary

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #519, opened 2023-06-29
Solution

pnakongit:


app/main.py
```diff
@@ -1 +1,55 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    FrontendDeveloper, BackendDeveloper, SoftwareEngineer
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
ArsenPidhoretskyi, line 50: This is not needed. It will do this automatically

app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.skills.extend(("Java", "Android studio"))
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    FrontendDeveloper, BackendDeveloper, SoftwareEngineer
```
ArsenPidhoretskyi, line 47:
```suggestion
    FrontendDeveloper, BackendDeveloper
```
FrontendDeveloper and BackendDeveloper already inherit from SoftwareEngineer

ArsenPidhoretskyi (changes requested):

app/main.py
pnakongit: Noted

vsmutok (approved):
Great 👍 

But pay attention to the empty lines after creating the class, you either leave them or delete them, but it needs to be the same everywhere

PR #520, opened 2023-07-01
engineer_1

CeO-PsH:


app/main.py
```diff
@@ -1 +1,54 @@
...
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
```
Polyakiv-Andrey, line 46: This class does not need an initialiser

app/main.py
```diff
@@ -1 +1,54 @@
...
+        super().__init__(name)
+        self.name = name
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
```
Polyakiv-Andrey, line 53: methods of inherited classes are called via self, and super is needed to override their behavior if necessary

app/main.py
```diff
@@ -1 +1,54 @@
...
+        super().create_awesome_web_page()
```
Polyakiv-Andrey, line 54: methods of inherited classes are called via self, and super is needed to override their behavior if necessary

Polyakiv-Andrey (changes requested):

app/main.py
CeO-PsH: How can i make default skills without initialiser, FullStackDeveloper  take skills from BackendDeveloper

Polyakiv-Andrey (approved):
Good job!

PR #524, opened 2023-07-04
Solution

anastmishchuk:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
ArsenPidhoretskyi, line 42: This is redundant code.

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        FullStackDeveloper.create_powerful_api(self)
+        FullStackDeveloper.create_awesome_web_page(self)
```
ArsenPidhoretskyi, line 47: just use 
self.create_powerful_api()

ArsenPidhoretskyi (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (approved):

PR #528, opened 2023-07-06
Solution

Anastasia-Su:


app/main.py
```diff
@@ -1 +1,54 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
ArsenPidhoretskyi, line 2: > Also, engineer should have skills attribute

Nothing was said about class attribute. It must be instance attribute

ArsenPidhoretskyi (changes requested):

Anastasia-Su:
Fixed. But could you please explain why instance attribute is better here than class attribute? Now FullStackDeveloper init method has instances created - seems less pretty than before.

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
vasylhnatiuk, line 14: Use += it will change list but not create new. Read readme.md and checklist also, change here and all of the code

vasylhnatiuk: It will help remove redundant code inside FullStackDeveloper

app/main.py
```diff
@@ -1 +1,52 @@
...
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 42: resolve previous comments and delete init from Full Stack Developer

vasylhnatiuk (changes requested):
You must not use class attribute because class attribute saves in the parent class, and if we change skills in one instance, all of the instances will change skills too.

vsmutok (approved):
Good Job 👍

PR #530, opened 2023-07-07
solution

vkleshko:


app/main.py
```diff
@@ -1 +1,52 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
ArsenPidhoretskyi, line 2: It must be instance attribute. Fix in all places

ArsenPidhoretskyi (changes requested):

PR #539, opened 2023-07-18
Implemented Py Software Engineers

mtiunov:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    FrontendDeveloper,
+    BackendDeveloper,
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Bohdan-Salamakha, line 46: u don't need init method here

Bohdan-Salamakha (changes requested):

app/main.py
mtiunov: Thank you

Bohdan-Salamakha (approved):

PR #545, opened 2023-07-23
solution

H1kki:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.default_skills = ["Java", "Android studio"]
```
Dimosphen1, line 35: There is no need for the `self.default_skills` attribute, you should extend `skills` with values directly. Make it here and in other classes as well.

Dimosphen1 (changes requested):
Several minor changes were requested.

pavlejviki (approved):
Good job!

PR #559, opened 2023-08-05
solution

EvgenMilokumov:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        super().__init__(name)
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 43: init is redundant here

app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = None
```
vasylhnatiuk, line 4:
```suggestion
        self.skills = []
```

app/main.py
```diff
@@ -1 +1,49 @@
...
+        if self.skills is None:
```
vasylhnatiuk, line 5: why you need it?

vasylhnatiuk (changes requested):

app/main.py
EvgenMilokumov: done

app/main.py
EvgenMilokumov: clear, thanks

PR #560, opened 2023-08-08
Develop

ostboiko:


app/main.py
```diff
@@ -1 +1,48 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_web_application(self) -> None:
```
YukihiroSM, line 44: Return annotation is wrong. YOu are returning not None

YukihiroSM (changes requested):

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Bohdan-Salamakha, line 42: u don't need an init method here

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #565, opened 2023-08-11
Implemented software engineers classes

KristianYe:


app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
```
Kroll410, line 7: Check for duplicates

Kroll410, line 7: Check for duplicates

Kroll410 (changes requested):

app/main.py
nyanplague: @Kroll410 Thank you!

Kroll410 (approved):

PR #576, opened 2023-08-13
Solution

denys-source:


app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name: str = name
+        self.skills: list[str] = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
```
Kroll410, line 7: Check duplicates

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
```
Kroll410, line 40: Set some skills for fullstack developer

Kroll410 (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += ["Java", "Android studio"]
```
Kroll410, line 34: You have to call learn_skill method from the base class

Kroll410 (changes requested):

PR #577, opened 2023-08-13
'Solution'

OstapIvakh:


app/main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
```
Kroll410, line 8: Check for duplicates

Kroll410 (changes requested):

Kroll410 (approved):
This solution works, but you could simply call learn_skill method from base class

PR #580, opened 2023-08-14
Solution

OleksandrMoroz24:


Kroll410 (changes requested):
Reuse `learn_skill` method from base class

PR #581, opened 2023-08-14
Solution

CaCtUs1k:


Kroll410 (changes requested):
Use learn_skill from base class

Kroll410 (commented):
Sorry, mistake in review :)

Kroll410 (approved):
GJ

PR #583, opened 2023-08-14
Solution with Inheritance

ZaichenkoViktoriia:


Kroll410 (changes requested):
Use learn_skill from base class in inits

Kroll410 (approved):
Sorry, mistake in review :)

PR #585, opened 2023-08-14
'Solution'

WonnaKoo:


Kroll410 (changes requested):
Not reassign the skills, but populate them, you can use extend method

Kroll410 (approved):

PR #587, opened 2023-08-14
'Solution'

grungezz:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
```
Kroll410, line 13: use for loops or extend method

Kroll410: here and after

Kroll410 (changes requested):
GJ, update one change

Kroll410 (approved):

PR #589, opened 2023-08-14
py-software-engineers

Vladyslav-Pliashechnyk:


Kroll410 (changes requested):
You don't need to re-assign skills - use extend or learn_skill method

PR #594, opened 2023-08-17
Solution

olgierrd:


app/main.py
```diff
@@ -1 +1,55 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
```
Polyakiv-Andrey, line 41: rewrite the initialization of the other methods so you don't have to write the initialization here

app/main.py
```diff
@@ -1 +1,48 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
pavlejviki, line 43: __init__ method is redundant as you just call super() with no custom logic

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #605, opened 2023-08-29
engineers_solution

Filllick:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().__init__(name=name)
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 42: init is useless here

vasylhnatiuk (changes requested):

Filllick:
init is useless because FullStackDeveloper inherits  whole __init__ method from both  BackendDeveloper and FrontendDeveloper?

vasylhnatiuk (approved):
yes, you are right

PR #612, opened 2023-09-03
Solution

AndriiYe:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Bohdan-Salamakha, line 42: u don't need an init method here

Bohdan-Salamakha (changes requested):

app/main.py
AndriiYe: yea it is got __init__ from parent classes

Dimosphen1 (approved):
LGTM, well done!

PR #615, opened 2023-09-03
Solution-py-software-engineers

Nadiia-developer:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "CSS", "HTML"]
```
Bohdan-Salamakha, line 43: u don't need an init method here

Bohdan-Salamakha (changes requested):
don't hardcode skills
u need to extend skills list in every class or use + operator to do it

Bohdan-Salamakha (approved):

PR #617, opened 2023-09-04
'Solution-software-engineers'

elenkomar:


app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self,
+                 name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self,
+                    skill: str) -> None:
+        self.skills += [skill]
```
ArsenPidhoretskyi, line 9: Use .append method

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self,
```
vasylhnatiuk, line 2:
```suggestion
    def __init__(self, name: str) -> None:
```

vasylhnatiuk: it would be better

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #621, opened 2023-09-06
With problem

Dmitriy-Poplinski:


app/main.py
```diff
@@ -1 +1,52 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str, skills: list | None = None) -> None:
```
Dimosphen1, line 2: You do not need to pass `skills` attribute here and in other classes as well.

app/main.py
```diff
@@ -1 +1,52 @@
...
+        self.name = name
+        if skills is None:
```
Dimosphen1, line 4: Do not use if condition here, just assign `self.skills` to an empty list.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #630, opened 2023-09-17
'Solution'

nsdNite:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Bohdan-Salamakha, line 46: u don't need an init method here

Bohdan-Salamakha (changes requested):

app/main.py
nsdNite: updated
thanks for the comment

Dimosphen1 (approved):
LGTM, well done!

PR #632, opened 2023-09-19
Solution

Andru37:


app/main.py
```diff
@@ -1 +1,67 @@
...
+        print(self.skills)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+full_stack_dev = FullStackDeveloper("Tom")
+sa = AndroidDeveloper("Sam")
```
YukihiroSM, line 67: This should be removed as it's not a part of module.

app/main.py
```diff
@@ -1 +1,67 @@
...
+
+
+class BackendDeveloper(SoftwareEngineer):
+
+    def __init__(self,
+                 name: str) -> None:
+        super().__init__(name)
+
+        for i in ["Python", "SQL", "Django"]:
+            self.learn_skill(i)
```
YukihiroSM, line 32: JFYI: you can use list.extend() here instead of `learn_skill`.

YukihiroSM (changes requested):

app/main.py
Andru37: Ok

app/main.py
Andru37: What practice is better?

YukihiroSM: I think in this case it's better to use extend, because you are hardcoding the values anyway. It will just look better.

YukihiroSM (approved):

PR #636, opened 2023-09-21
Solution

MatsalakO:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
viktoria-rybenchuk, line 42: you can remove the init method in this class, as all the necessary skills will be inherited from the parent classes automatically

viktoria-rybenchuk (changes requested):

Bohdan-Salamakha (approved):

PR #637, opened 2023-09-21
Develop

asdadaversa:


app/main.py
```diff
@@ -1 +1,62 @@
...
+class SoftwareEngineer:
+
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        for super_skill in skill.split():
+            self.skills.append(super_skill)
```
viktoria-rybenchuk, line 13: avoid using for loop

app/main.py
```diff
@@ -1 +1,62 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    skills = ["JavaScript", "HTML", "CSS"]
```
viktoria-rybenchuk, line 18: read checklist and fix in alll places

app/main.py
```diff
@@ -1 +1,62 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
```
viktoria-rybenchuk, line 58: this is redundant code, you can remove the init method in this class, as all the necessary skills will be inherited from the parent classes automatically

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,59 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+        self.skills += []
```
Bohdan-Salamakha, line 54: you don't need an init method here

app/main.py
asdadaversa: when i remove here, tests don't pass

Bohdan-Salamakha (changes requested):

app/main.py
asdadaversa: show please how, when i remove here, tests don't pass

app/main.py
asdadaversa: done))

Bohdan-Salamakha (approved):

PR #645, opened 2023-10-02
solution

Familenko:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        for skill in ["JavaScript", "HTML", "CSS"]:
+            self.learn_skill(skill)
```
LiudmylaKulinchenko, line 14: Here u can do like this self.skills += ["JavaScript", "HTML", "CSS"] or use `extend` to write less code

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+
```
LiudmylaKulinchenko, line 46: No need to rewrite init method here

LiudmylaKulinchenko (changes requested):

app/main.py
Familenko: I just wanted to use method self.learn_skill from SoftwareEngineer to show power of inheritance. But this method is taking one argument at a time

app/main.py
Familenko: I make too much copy paste)))

LiudmylaKulinchenko (approved):
Well done 😸

PR #653, opened 2023-10-07
Solution for py_software_engineers

wataschi:


app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    skills = ["JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = FrontendDeveloper.skills
```
viktoria-rybenchuk, line 17: Use the extend() method or += operator to add multiple values to a single list. Fix in all places

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = BackendDeveloper.skills + FrontendDeveloper.skills
```
viktoria-rybenchuk, line 50: This is redundant. You have already inherited this skill from `BackendDeveloper` and `FrontendDeveloper`

viktoria-rybenchuk (changes requested):

app/main.py
Dimosphen1, line 50: Do not modify the `__init__` method in the `FullStackDeveloper` class, you could omit it and relay on the initialization of parent classes to populate skills.

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
Dimosphen1, line 40: You should not redefine the `self.skills` attribute but rather extend the one that is defined in the `SoftwareEngineer` class. In this way, you will make an in-place list modification that is expected here.

app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.skills = FrontendDeveloper.skills
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+
+    skills = ["Python", "SQL", "Django"]
```
Dimosphen1, line 26: Delete class attributes here and for `FrontendDeveloper`, extend the `self.skills` directly with the required values.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
GJ!

PR #662, opened 2023-10-09
py-software-engineers Solution

RVChornyy:


app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
mykolamateichuk, line 2: `skills` should be initialized in `__init__` method

mykolamateichuk (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
LiudmylaKulinchenko, line 47: No need to rewrite init method here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Stunning 🌼

PR #663, opened 2023-10-09
Solution

mrve4nick:


app/main.py
```diff
@@ -1 +1,57 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
pavlejviki, line 2: skills should be an attribute of the instance not the class

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
pavlejviki, line 12: Use the extend() method or += operator to add multiple values to a single list.(see checklist)

FrontendDeveloper will inherit an empty skills  list from its parent, you need to extend that list instead of creating a new one

app/main.py
```diff
@@ -1 +1,57 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    skills = FrontendDeveloper.skills + BackendDeveloper.skills
+
+    def __init__(self, name: str) -> None:
```
pavlejviki, line 50: if you follow the comments above  the def __init__ method in FullStackDeveloper class will become redundant as all the necessary skills gonna be inherited from its parents

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
pavlejviki, line 45: You can remove the init method in this class, its redundant as just calls super().__init__

pavlejviki (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #667, opened 2023-10-10
Solution

katyakoziar:


OlehShumov (approved):

app/main.py
```diff
@@ -1 +1,54 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
LiudmylaKulinchenko, line 2: skills variable should be an attribute of the instance, not class

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
LiudmylaKulinchenko, line 13: Use += operator or extend() to extend a list instead of redefining it

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
fsocie7y, line 41: No need to rewrite init method here

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🔥

PR #669, opened 2023-10-10
Solution

Shatrovskyi:


app/main.py
```diff
@@ -1 +1,62 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "Python",
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
```
Andrii-Silenko, line 57: I think we have to find the way how to ineritance this skills from parents' classes, not just assign this list to self.skills. But maybe I'm wrong)

Andrii-Silenko (approved):

app/main.py
Shatrovskyi: Yep, You are right.

LiudmylaKulinchenko (approved):
Well done 😸

PR #672, opened 2023-10-10
Solution

Andrii-Silenko:


RabenkoYevhenii (approved):

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for skill in ["JavaScript", "HTML", "CSS"]:
+            super().learn_skill(skill)
```
LiudmylaKulinchenko, line 14: Here u can do like this self.skills += ["JavaScript", "HTML", "CSS"] or use `extend` to write less code

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
LiudmylaKulinchenko, line 45: No need to rewrite init method here, it does nothing here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! 🐣

PR #674, opened 2023-10-10
Implemented all the following classes and methods!

raychw:
Implemented all the following classes and methods!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,57 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,57 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    BackendDeveloper,
+    FrontendDeveloper,
+    SoftwareEngineer
```
Dimosphen1, line 47: Do not inherit from `SoftwareEngineer`, its methods would be available because both `BackendDeveloper` and `FrontendDeveloper` inherit from it.

app/main.py
```diff
@@ -1 +1,57 @@
...
+):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Dimosphen1, line 51: Do not use `__init__` for this class, rely on the implementation of the parent classes.

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done!

PR #681, opened 2023-10-10
Solution Py Software Engineers

Oxbay:
it was hard for me)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,44 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,44 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += "JavaScript", "HTML", "CSS"
```
Wonsky1, line 13: maybe here would be better to be not "JavaScript", "HTML", "CSS", but specifically list

Wonsky1 (approved):

app/main.py
LiudmylaKulinchenko: I agree, then it won't confuse anybody and will be more readable

LiudmylaKulinchenko (changes requested):

PR #684, opened 2023-10-11
Solution

dkibalenko:


app/main.py
```diff
@@ -1 +1,81 @@
...
+        """
+        Displays an info message and returns the UX of the created app.
+        :return: str
+        """
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
Polyakiv-Andrey, line 71: inti should be omitted in this class

Polyakiv-Andrey (changes requested):

app/main.py
dkibalenko: sure, thanks!

viktoria-rybenchuk (approved):
GJ!

PR #685, opened 2023-10-11
Solution

mishagitcode:


app/main.py
```diff
@@ -1 +1,53 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
YukihiroSM, line 2: Skills should be object attribute, not class attribute. It should be created in init. Fix everywhere

app/main.py
```diff
@@ -1 +1,53 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
+
```
YukihiroSM, line 13: You should extend skills, not reset them. Also take a look into previous comment.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.append("JavaScript")
+        self.skills.append("HTML")
+        self.skills.append("CSS")
```
YukihiroSM, line 15:
```suggestion
        self.skills.extend(["JavaScript", "HTML", "CSS"])
```

YukihiroSM: Fix everywhere.

YukihiroSM (changes requested):

YukihiroSM (approved):
Good job!

PR #687, opened 2023-10-13
Solution

sixsixsixsamurai:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
pavlejviki, line 13: Use += operator or extend() to extend a list instead of redefining it

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
pavlejviki, line 43: if you follow the comment above you can remove __init__ from this class, as all necessary skills will be inherited from the parent classes

pavlejviki (changes requested):

vasylhnatiuk (approved):

PR #688, opened 2023-10-13
solution

oleksashcherbakov:


oleksashcherbakov:
please check

app/main.py
```diff
@@ -1 +1,54 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
pavlejviki, line 2: skills should be an attribute of the instance not the class

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
pavlejviki, line 13: Use += operator or extend() to extend a list instead of redefining it

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
```
pavlejviki, line 48: if you follow the comment above you can remove init from this class, as all necessary skills will be inherited from the parent classes

pavlejviki (changes requested):

oleksashcherbakov:
fixed

pavlejviki (approved):
Good job!

PR #689, opened 2023-10-13
Solution

frezworx:


app/main.py
```diff
@@ -1 +1,66 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = self.skills if self.skills else [
```
pavlejviki, line 15: self.skills is an empty list inherited from its parent. you just need to extend the list with the certain skill related to a particular class. 

Use += operator or extend() to extend a list instead of redefining it

app/main.py
```diff
@@ -1 +1,66 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+        fd_skills = FrontendDeveloper(self.name).skills
+        self.skills.extend(fd_skills)
```
pavlejviki, line 61: if you follow the comment above you can remove init from this class, as all necessary skills will be inherited from the parent classes

pavlejviki (changes requested):

pavlejviki (changes requested):
Please make changes according to the comments

pavlejviki (changes requested):
Please make changes according to the comments

PR #697, opened 2023-10-18
Solution

paashkovaaa:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
YukihiroSM, line 13: use self.skills.extend() insead of setting skills directly

YukihiroSM: Fix everywhere

YukihiroSM (changes requested):

PR #698, opened 2023-10-18
Solution

ibulakevych:


app/main.py
```diff
@@ -1 +1,63 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
+        self.skills = ["JavaScript",
+                       "HTML",
+                       "CSS"
+                       ]
```
YukihiroSM, line 17: Use skills.extend(...)
What if some skills will be adeded by default to SoftwareEngineer?

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,61 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper,
+                         FrontendDeveloper,
+                         SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
```
ArsenPidhoretskyi, line 56: This code is redundant

app/main.py
```diff
@@ -1 +1,61 @@
...
+                            ])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper,
+                         FrontendDeveloper,
+                         SoftwareEngineer):
```
ArsenPidhoretskyi, line 53: Why do you need inherit from SoftwareEngineer

ArsenPidhoretskyi (changes requested):

app/main.py
EdAlekseiev: Still not fixed

app/main.py
```diff
@@ -1 +1,58 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper,
+                         FrontendDeveloper
+                         ):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
```
EdAlekseiev, line 53:
```suggestion
```

EdAlekseiev (approved):
Good job

PR #702, opened 2023-10-21
Solution

DenVia:


app/main.py
```diff
@@ -1 +1,54 @@
...
+        self.learn_skill("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
LiudmylaKulinchenko, line 49: No need to rewrite init method here

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
+        self.learn_skill("HTML")
+        self.learn_skill("CSS")
```
LiudmylaKulinchenko, line 17: Here u can do like this self.skills += ["JavaScript", "HTML", "CSS"] or use `extend` to write less code

LiudmylaKulinchenko (approved):
Fantastic! 🌼

PR #709, opened 2023-10-29
solution

Andrey1104:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    FrontendDeveloper,
+    BackendDeveloper
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS", "Python", "SQL", "Django"]
```
Dimosphen1, line 46: Omit using `__init__` in this class, you could rely only on the `__init__` of the parent classes to satisfy requirements and populate `self.skills`.

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
Dimosphen1, line 33: Do not redefine `self.skills` in each subclass of `SoftwareEngineer` class, you should extend the list that was created in the `SoftwareEngineer` class with new values instead.

app/main.py
```diff
@@ -1 +1,51 @@
...
+    BackendDeveloper
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS", "Python", "SQL", "Django"]
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
Dimosphen1, line 51: Do not call these methods from `super()`, use `self` instead.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    FrontendDeveloper,
+    BackendDeveloper
+):
+    skills = []
```
Dimosphen1, line 49: Do not declare a class variable `skills`, there is no need for this.

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
+        self.learn_skill("HTML")
+        self.learn_skill("CSS")
```
Dimosphen1, line 15: Instead of calling `learn_skill` several times, you could extend the `self.skills` directly, e.g.
```suggestion
        self.skills.extend(["JavaScript", "HTML", "CSS"])
```
Make this change in all classes.

Dimosphen1 (changes requested):

app/main.py
Andrey1104: of course) forgot to remove it

app/main.py
Andrey1104: what about pull request "knights_fighting"???

Andrey1104: ok, why do we need a function learn_skill?

Dimosphen1 (approved):
LGTM (looks good to me), well done!

PR #711, opened 2023-10-30
Solution

maxvell4:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
pavlejviki, line 43: there is no need to declare __init__ just to call super(). it will be called automatically if the subclass does not have an __init__ method of its own.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #714, opened 2023-11-02
'Solution'

yul1valko:


app/main.py
```diff
@@ -1 +1,57 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
pavlejviki, line 12: no need to add it as a class attribute, add straight to the instance attribute

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = FrontendDeveloper.skills
```
pavlejviki, line 16: Use the extend() method or += operator to add multiple values to a single list.

FrontendDeveloper inherited an empty skills list from SoftwareEngineer, extend it instead of overriding.

app/main.py
```diff
@@ -1 +1,57 @@
...
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = ["Python", "SQL", "Django"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = BackendDeveloper.skills
```
pavlejviki, line 28: see comments above

app/main.py
```diff
@@ -1 +1,57 @@
...
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    skills = ["Java", "Android studio"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = AndroidDeveloper.skills
```
pavlejviki, line 40: see comments above

app/main.py
```diff
@@ -1 +1,57 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    skills = BackendDeveloper.skills + FrontendDeveloper.skills
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = FullStackDeveloper.skills
```
pavlejviki, line 52: after fixing the comments above the __init__ method in this class will become redundant as all necessary skills will be inherited from parent classes automatically

pavlejviki (changes requested):

yul1valko:
Have implemented requested changes

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
pavlejviki, line 47: you can remove __init__ as it just calls super() without any custom logic

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #724, opened 2023-11-09
Software_engineers_solution

MykolaMazh:


app/main.py
```diff
@@ -1 +1,63 @@
...
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(
+            self,
+            name: str
+    ) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
viktoria-rybenchuk, line 19: Use the `extend()` method or `+=` operator to add multiple values to a single list.

viktoria-rybenchuk: Fix in all places

app/main.py
```diff
@@ -1 +1,63 @@
...
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(
+            self,
+            name: str
+    ) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+
```
viktoria-rybenchuk, line 59: Skills are inherited from parent classes, no need to set them again.

viktoria-rybenchuk (changes requested):

app/main.py
MykolaMazh: image
image


image
image


image


image
image


in this case i understand: MRO, inheritance of  BackendDeveloper ```__init__```

**but WHY!**
image
image
image
image

**WHY ["JavaScript", "HTML", "CSS"] come first**

if both ```__init__``` methods of father classes launch **WHY** second ```super().__init__ ```  in FrontendDeveloper **doesn't erase** 
```python 
["Python", "SQL", "Django"]
```
**to**
```python
self.skills = []
```

viktoria-rybenchuk: Come to Q&A with this question

viktoria-rybenchuk (approved):
GJ!

PR #731, opened 2023-11-14
solution

boris-devs:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for skill in "JavaScript", "HTML", "CSS":
+            self.skills.append(skill)
```
EdAlekseiev, line 14: Use `extend` method instead of append element in cycle

app/main.py
```diff
@@ -1 +1,53 @@
...
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for skill in "Python", "SQL", "Django":
+            self.skills.append(skill)
```
EdAlekseiev, line 25: The same comment as above

app/main.py
```diff
@@ -1 +1,53 @@
...
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for skill in "Java", "Android studio":
+            self.skills.append(skill)
```
EdAlekseiev, line 36: The same comment as above

app/main.py
```diff
@@ -1 +1,53 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
EdAlekseiev, line 47: This method is redundant. You should delete this method

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.extend([skill])
```
YukihiroSM, line 7: You are adding one skill here. Using of extend method and transforming skill into list is not needed here. 
```suggestion
    def learn_skill(self, skill: str) -> None:
        self.skills.append(skill)
```

YukihiroSM (changes requested):

YukihiroSM (approved):
Nicely done!

PR #732, opened 2023-11-15
Solution

Serhii-Hrynyshyn:


app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
YukihiroSM, line 15: skills should be instance attribute, not class. Add it to init. Also, use extend method, please. 
Fix it in all classes

app/main.py
```diff
@@ -1 +1,54 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(AndroidDeveloper,
+                         BackendDeveloper,
+                         FrontendDeveloper,
+                         SoftwareEngineer,
+                         ):
+    skills = (FrontendDeveloper.skills + BackendDeveloper.skills)
```
YukihiroSM, line 49: This should be removed after fixing previous comments.

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(AndroidDeveloper,
```
YukihiroSM, line 44: FullStack developer shouldn't be Android developer. Fullstack is front + back.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,60 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
EdAlekseiev, line 15: You need to extend `self.skills` instead of rewrite.
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

app/main.py
```diff
@@ -1 +1,60 @@
...
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
EdAlekseiev, line 26: The same comment as above

app/main.py
```diff
@@ -1 +1,60 @@
...
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
EdAlekseiev, line 37: The same comment as above

app/main.py
```diff
@@ -1 +1,60 @@
...
+                         ):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python",
+                       "SQL",
+                       "Django",
+                       "JavaScript",
+                       "HTML",
+                       "CSS",
+                       ]
```
EdAlekseiev, line 55: This method is redundant

EdAlekseiev (changes requested):

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #733, opened 2023-11-15
Solution

alex329657:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
ArsenPidhoretskyi, line 14: Use extend or += to modify skills instead of override

ArsenPidhoretskyi: Fx in all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #740, opened 2023-11-22
Solution

vpanchenkoSPS:


app/main.py
```diff
@@ -1 +1,53 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
+        self.learn_skill("HTML")
+        self.learn_skill("CSS")
```
fsocie7y, line 16: Here u can do like this self.skills += ["JavaScript", "HTML", "CSS"] or use `extend` to write less code

fsocie7y: same in other similar cases

app/main.py
```diff
@@ -1 +1,53 @@
...
+        self.learn_skill("Java")
+        self.learn_skill("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
fsocie7y, line 47: No need to rewrite init method here

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #751, opened 2023-11-30
Develop

Maxboond:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for skill in ["JavaScript", "HTML", "CSS"]:
+            self.learn_skill(skill)
```
ArsenPidhoretskyi, line 17: Use .extend or += instead of loop
Fix in all places

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
ArsenPidhoretskyi, line 48: This is redundant. It will automatically do the same

ArsenPidhoretskyi (changes requested):

pavlejviki (approved):
Good job!

PR #759, opened 2023-12-05
solve py-software-engineers

zdmytryszyn:


app/main.py
```diff
@@ -1 +1,58 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
pavlejviki, line 2: skills should be an instance attribute, not a class attribute

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
pavlejviki, line 12: Use the extend() method or += operator to add new skills to the empty skills list inherited from the parent class

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = ["Python", "SQL", "Django"]
```
pavlejviki, line 23: see comment above

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    skills = ["Java", "Android studio"]
```
pavlejviki, line 34: see comment above

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+class FullStackDeveloper(
+    BackendDeveloper,
+    FrontendDeveloper,
+    SoftwareEngineer
+):
+    skills = BackendDeveloper.skills + FrontendDeveloper.skills
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
pavlejviki, line 52: no need to declare init method if there is no custom logic for this class

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #763, opened 2023-12-07
Solution X1

yaroslavsysoiev:
I wrote the code, ran pytest and flake8 app. I checked checklist and didn't find of 'bad practice' but would appreciate any comments about.


viktoria-rybenchuk (approved):
Good job!

PR #766, opened 2023-12-12
solution

AlisaSkl:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += ["Python", "SQL", "Django"]
```
Dimosphen1, line 46: You do not need to declare `__init__` in this class, rely completely on the parent classes implementation.

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
Dimosphen1, line 15: Instead of redefining the `self.skills` list defined in the `SoftwareEngineer` class, extend `self.skills` from the `SoftwareEngineer` class with new values so you would not create a new list of skills in each class. Change the `__init__` of other classes that add new skills as well.

Dimosphen1 (changes requested):

PR #767, opened 2023-12-13
Solution

Artkom98:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
YukihiroSM, line 12: If SoftwareEngineer will have some skills - you will overwrite it instead of extending. 
Use skills.extend() here and in other classes

tests/test_main.py
```diff
@@ -1,4 +1,5 @@
 import io
+import io
 from contextlib import redirect_stdout
```
YukihiroSM, line 4: Revert changes here, please.

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #773, opened 2023-12-20
Solution

M4Xpy:


app/main.py
```diff
@@ -1 +1,69 @@
...
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
```
NikOneZ1, line 19: Use the extend() method or += operator to add multiple values to a single list

app/main.py
```diff
@@ -1 +1,69 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "Python",
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
```
NikOneZ1, line 64: You don't need to redefine skills here, as it's already defined in the parent classes

NikOneZ1 (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #774, opened 2023-12-20
Solution

halyna-baklanova:


app/main.py
```diff
@@ -1 +1,53 @@
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.extend([skill])
```
pavlejviki, line 7: better to use append as its gonna be just one skill

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
+        self.learn_skill("HTML")
+        self.learn_skill("CSS")
```
pavlejviki, line 15: instead of calling the method you can straight extend the skills list. see checklist how

fix in all of the classes

app/main.py
```diff
@@ -1 +1,53 @@
...
+        self.learn_skill("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
pavlejviki, line 47: if there is no custom logic added to init there is no need to declare it

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
 # write your code here
```
pavlejviki, line 53: remove comment

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #783, opened 2023-12-28
Solution

msymonovych:


aLEKS-e3 (approved):

aLEKS-e3:
the only issue is that the _learn_skill_ function takes a _skill_ of **string** type, but you add a **list** of functions in FullStack and Backend classes.

PR #793, opened 2024-01-02
Solution

TymurSh:


NikOneZ1 (changes requested):
It would be better to define skills in init method

TymurSh:
> It would be better to define skills in init method
Isn't it necessary to create skills as a class attribute in the task conditions? I can't pass the last test when skills is an instance attributes

app/main.py
```diff
@@ -1 +1,65 @@
-# write your code here
```
TymurSh, line 1: Isn't it necessary to create skills as a class attribute in the task conditions? I can't pass the last test when skills is an instance attributes

TymurSh:
Isn't it necessary to create skills as a class attribute in the task conditions? I can't pass the last test when skills is an instance attributes

TymurSh (commented):
Isn't it necessary to create skills as a class attribute in the task conditions? I can't pass the last test when skills is an instance attributes

PR #803, opened 2024-01-08
Solution

annaminaieva:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Dimosphen1, line 42: There is no need for this method, remove it since `__init__` will be inherited from the parent classes.

Dimosphen1 (changes requested):
Several minor changes were requested.

pavlejviki (approved):
Good job!

PR #821, opened 2024-01-17
Solution

VladKyianytsia:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
barabarinov, line 42: You can even remove __init__ from here and it will works!

barabarinov (changes requested):

app/main.py
VladKyianytsia: Agree

PR #824, opened 2024-01-17
Task solution

shaiduchyk:


Serejka21 (approved):
AWESOME! Best solution what i ever seen

PR #841, opened 2024-01-23
'Solution'

sofiazholud:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 44: you can avoid init in this class

vasylhnatiuk (changes requested):

app/main.py
sofiazholud: rewrited it

EdAlekseiev (approved):
Good job

PR #842, opened 2024-01-25
Solution

AlexGrytsai:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+        return self
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
vasylhnatiuk, line 17:
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

vasylhnatiuk: you must add to previous skills

app/main.py
```diff
@@ -1 +1,53 @@
...
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 45: if you use += you can avoid `init` here

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
.

YukihiroSM (approved):
Looks good! Also, note, that you can use something like: self.skills.extend() to extend existing list with some values.

PR #847, opened 2024-01-27
Solution

mkuchuman:


app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

vasylhnatiuk: don't replace, but add for prevoius skills

app/main.py
```diff
@@ -1 +1,55 @@
...
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self, ) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 41: and if you add you don't need init in this class

vasylhnatiuk (changes requested):

mkuchuman:
thank you, I didn’t even think that it was possible to implement code like this :)

EdAlekseiev (approved):
Good job

PR #861, opened 2024-02-04
Solution

MaksymNikulyak:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name=name)
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 41: this init is useless here

vasylhnatiuk (changes requested):

app/main.py
MaksymNikulyak: Indeed so, thank you!

vasylhnatiuk (approved):

PR #869, opened 2024-02-08
solved task

Shantagust:


app/main.py
```diff
@@ -1 +1,51 @@
...
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        super().learn_skill("Java")
+        super().learn_skill("Android studio")
```
pavlejviki, line 37: why do you call the method twice? why don't you follow the previous logic?

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
```
pavlejviki, line 46: no need to declare __init__ without any customisation

pavlejviki (changes requested):

app/main.py
Shantagust: agree, my bad

pavlejviki (approved):
Well done

PR #871, opened 2024-02-08
'Solution'

liudmylaperchevska:


app/main.py
```diff
@@ -1 +1,48 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, *skill: str) -> None:
+        self.skills.extend(skill)
```
pavlejviki, line 7: you can use append when you need to add just one element

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
pavlejviki, line 13: use the extend() method or += operator to add multiple skills to the already existing self.skills list inherited from the parent class

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
pavlejviki, line 23: see comment above

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
pavlejviki, line 33: see comment above

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
pavlejviki, line 43: after making all the necessary changes this __init__ will be redundant

pavlejviki (changes requested):

PR #879, opened 2024-02-14
Add Solution

CreepyMo:


app/main.py
```diff
@@ -1 +1,59 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
pavlejviki, line 2: skills should be an instance attribute

app/main.py
```diff
@@ -1 +1,59 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
pavlejviki, line 12: instead of creating a new list of skills, extend the inherited one from the parent class

see checklist

app/main.py
```diff
@@ -1 +1,59 @@
...
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+full_stack_dev = FullStackDeveloper("Tom")
+print(full_stack_dev.skills)
+
+full_stack_dev.create_web_application()
```
pavlejviki, line 59: remove redundant code

pavlejviki (changes requested):

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Dimosphen1, line 47: There is no need to place `__init__` here, you could rely on the `__init__` of the parent classes. Remove it.

Dimosphen1 (changes requested):

app/main.py
CreepyMo: Done

Dimosphen1 (approved):
LGTM, well done!

PR #892, opened 2024-02-22
Solution

twonumberfortyfives:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
Polyakiv-Andrey, line 41: You should rewrite parent classes so that in this class you don't need to use an initializer

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+class AndroidDeveloper(SoftwareEngineer):
+    skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    skills = FrontendDeveloper.skills + BackendDeveloper.skills
```
Polyakiv-Andrey, line 37: You should use inheritance to make sure that you don’t have to write Init in the FullStackDeveloper class. And in all other classes it is necessary.

Polyakiv-Andrey (changes requested):

app/main.py
twonumberfortyfives: Вибачте будь ласка, я не розумію, що треба зробити

twonumberfortyfives: > You should rewrite parent classes so that in this class you don't need to use an initializer

Вибачте будь ласка, я не розумію, що треба зробити

Polyakiv-Andrey (approved):
Good job!

PR #893, opened 2024-02-23
Solution

Voronessko:


Voronessko:
in___

        if isinstance(skill, str):
            self.skills.append(skill)
        else:
            self.skills.extend(skill)
            
why skill must be "str"? i think it shold be a list....  (but tests complane about it)

EdAlekseiev:
> in___
> 
> def learn_skill(self, skill: str) -> None: if isinstance(skill, str): self.skills.append(skill) else: self.skills.extend(skill)
> 
> why skill must be "str"? i think it shold be a list.... (but tests complane about it)

https://github.com/mate-academy/py-software-engineers/tree/master#:~:text=SoftwareEngineer%20class%20should%20have%20one%20method%3A%20learn_skill%20that%20takes%20a%20skill%20of%20string%20type%20and%20should%20add%20it%20to%20the%20skills%20list.

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.skills = []
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        if isinstance(skill, str):
+            self.skills.append(skill)
+        else:
+            self.skills.extend(skill)
```
EdAlekseiev, line 11: skill can be only `str`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,51 @@
...
+        for new_skill in ["Java", "Android studio"]:
+            self.learn_skill(new_skill)
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 45: you can avoid init in this class. delete it

vasylhnatiuk (changes requested):

PR #894, opened 2024-02-24
Develop

MikolaRozhansky:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
EdAlekseiev, line 13: You should specify skills inside method `__init__`. Fix everywhere.

EdAlekseiev (changes requested):

app/main.py
MikolaRozhansky: From readme.md
`Firstly, you should create SoftwareEngineer class, which will be a parent for all other classes. Its __init__ method should take one parameter: name of the engineer. Also, engineer should have skills attribute - a list of his/her skills, such as "Python" or "JavaScript". skills should be initialised with empty list.`

Іts __init__ method should take **one parameter**: name of the engineer.

MikolaRozhansky: Then you will get a discrepancy with the task

vasylhnatiuk: yes, and init you should add skills to exists skills

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
```suggestion
        super().__init__(name)
        skills += ["JavaScript", "HTML", "CSS"]
```

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #898, opened 2024-02-24
Develop

Rudge0:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

app/main.py
```diff
@@ -1 +1,53 @@
...
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 45: you don't need init here if you use += in parent classes

vasylhnatiuk: bacause all of skills will be inherited

app/main.py
vasylhnatiuk: do everywhere

vasylhnatiuk (changes requested):

PR #903, opened 2024-02-26
result

Olevsky-Ivan:


app/main.py
```diff
@@ -1 +1,56 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "Python",
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS"
+        ]
```
Dimosphen1, line 51: Do not define the `__init__` method in this class, the `skills` should be populated using the parent classes

app/main.py
```diff
@@ -1 +1,56 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
Dimosphen1, line 14: Instead of redefining the `self.skills` list from the parent class, add new values in place using `+=` or `extend`. Apply this change to all classes

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Dimosphen1, line 43: Remove the `__init__` from this class

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
Dimosphen1, line 24: Use `+=` here as well

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
Dimosphen1, line 34: Use `+=` here as well

Dimosphen1 (changes requested):
Requested changes were not implemented, revise previous comments

app/main.py
Olevsky-Ivan: is expected error

app/main.py
Olevsky-Ivan: and here
изображение_2024-02-27_123852445

Olevsky-Ivan: only Frontend +=

Dimosphen1: Inherit `AndroidDeveloper` from the `SoftwareEngineer` as mentioned in the task description and implement the change that was requested

app/main.py
Dimosphen1: Implement the change that was requested and make sure that inheritance would remain correct

Dimosphen1 (changes requested):

PR #935, opened 2024-03-16
Solution

Vova-Stoliar:


app/main.py
```diff
@@ -1 +1,76 @@
-# write your code here
+class SoftwareEngineer:
+    skills: list[str] = []
```
LiudmylaKulinchenko, line 2: skills variable should be an attribute of the instance, not class

app/main.py
```diff
@@ -1 +1,76 @@
...
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills: list[str] = [
+        "JavaScript",
+        "HTML",
+        "CSS"
+    ]
```
LiudmylaKulinchenko, line 17: Use += operator or extend() to extend a list instead of redefining it

app/main.py
```diff
@@ -1 +1,76 @@
...
+        "Python",
+        "SQL",
+        "Django",
+        "JavaScript",
+        "HTML",
+        "CSS",
+    ]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
LiudmylaKulinchenko, line 70: You won't need to rewrite an init method in this class after making changes

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,63 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
LiudmylaKulinchenko, line 14: Use += operator or extend() to extend a list instead of redefining it

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🔥

PR #947, opened 2024-03-21
Solution

DmytroHlazyrin:


panicua (approved):
We think alike our only difference is that I write in super method ```name=name```, you just ```name```, and I created list like ```list()```, you did it ```[]```. Or.. task doesn't let the creativity :D

Oleksl888 (approved):
Good job!

PR #954, opened 2024-03-23
Solution

y-kondrashova:


app/main.py
```diff
@@ -1 +1,59 @@
...
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = ["Python", "SQL", "Django"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
YuriiKindrat, line 27: same

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Oleksl888, line 43: You don't need to explicitly call __init__ here since you are not going to add any extra attributes

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_web_application(self) -> None:
+        backend_developer = BackendDeveloper
```
Oleksl888, line 46: You don't need to instantiate backend and frontend developers here. Since you inherit your class both from FrontendDeveloper, BackendDeveloper your class will have all attributes and methods from both those classes

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ!

PR #960, opened 2024-03-24
Solution#1

VitalyBashkiser:


semermaks (approved):
GJ

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
YuriiKindrat, line 42: why do u need this ?)

YuriiKindrat (changes requested):

app/main.py
VitalyBashkiser: We have added the def __init__/super() function, it really has nothing to do here

Oleksl888 (approved):
Good Job!

PR #967, opened 2024-03-25
Solution Software Engineers March 25 ver 1.0

DmytroPieskov:


app/main.py
```diff
@@ -1 +1,59 @@
...
+        super().__init__(name)
+        self.name = name
+        self.skills = [
+            "Python",
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
```
Dimosphen1, line 54: Do not define `__init__` in this class, rely on the implementation of the parent classes

app/main.py
```diff
@@ -1 +1,59 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
+        self.skills = ["JavaScript", "HTML", "CSS", ]
```
Dimosphen1, line 14: Do not redefine the list for `self.skills` defined in the parent class. Extend it with new values using `+=` or `.extend()` method instead. Apply this change to the `BackendDeveloper` and `AndroidDeveloper` classes as well.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.name = name
```
Dimosphen1, line 13: There is no need to redefine `name` since it is already performed in the parent class. Remove this line from all classes

Dimosphen1 (changes requested):

PR #970, opened 2024-03-25
Implement Py Software Engineers task

bwaah1:


mikhailyemets (approved):

app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
Oleksl888, line 13: Here you overwrite skills that may have been acquired by parent class. Check this in all your classes

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "CSS", "HTML"]
```
Oleksl888, line 43: You don't need to hard code this, skills should be inherited from parent classes. If you do this right, you will not need `__init__` method here at all

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
```
Oleksl888, line 47: since you inherit from these developers you must call these methods from FullStack class

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #971, opened 2024-03-25
Solution

Ruxxell54:


app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name: str = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
```
alonasorochynska, line 7: no need for condition

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        api_address = self.create_powerful_api()
+        webpage = self.create_awesome_web_page()
+        return api_address, webpage
```
alonasorochynska, line 49: this method have to return None

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
```
alonasorochynska, line 34: no need to extend here

app/main.py
CharmNiovi, line 49: Can be made without return.
_his method should print the following message: "{name} started creating a web application..." and call create_powerful_api, create_awesome_web_page methods._
Just a call

YuriiKindrat, line 49: no need to use return here, u could just call em

YuriiKindrat (approved):
LGTM

PR #972, opened 2024-03-25
'Solution'

afaidiuk:


siaserb (approved):

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
```
Oleksl888, line 14: Calling learn_skill 3 times is not optimal. Think what list methods you can use to make it more simple and readable. Fix in other methods as well

app/main.py
```diff
@@ -1 +1,53 @@
...
+        self.learn_skill("Java")
+        self.learn_skill("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
Oleksl888, line 47: Why do you need __init__ here?

Oleksl888 (changes requested):

PR #979, opened 2024-03-28
'Solution'

kssshevchenko:


app/main.py
```diff
@@ -1 +1,62 @@
...
+        if skill not in self.skills:
+            self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        default_skills = ["JavaScript", "HTML", "CSS"]
+        for skill in default_skills:
+            self.learn_skill(skill)
```
SerhiiMar, line 19: You don't have to add each skill through the `learn_skill()` method because there is an empty(!) `skills` attribute.
You can just add a whole list of default skills to `skills` attribute.

Same with `BackendDeveloper` and `AndroidDeveloper` classes.

app/main.py
```diff
@@ -1 +1,62 @@
...
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        SoftwareEngineer.__init__(self, name)
+        backend_skills = ["Python", "SQL", "Django"]
+        frontend_skills = ["JavaScript", "HTML", "CSS"]
+        all_skills = backend_skills + frontend_skills
+        for skill in all_skills:
+            self.learn_skill(skill)
```
SerhiiMar, line 57: This code is unnecessary. Delete it and nothing will change :)

SerhiiMar (changes requested):
Delete unnecessary code.

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.default_skills = ["Python", "SQL", "Django"]
+        self.skills.extend(self.default_skills)
```
YuriiKindrat, line 28: why do u need this attr, u can use extend for all skill in every class

YuriiKindrat (approved):

PR #1002, opened 2024-04-16
Solution

mlupekha:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 41: you can't avoid init here

vasylhnatiuk (changes requested):

app/main.py
mlupekha: > you can't avoid init here

what do you mean? it present here

mlupekha: > you can't avoid init here

@vasylhnatiuk, I'm not getting it. Isn't init present here now?

EdAlekseiev (changes requested):

PR #1003, opened 2024-04-16
all tests pass

aniashev:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 41: this class should inherit skills from parent classes, you don't need init here

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: delete init from FullStackDeveloper

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1004, opened 2024-04-16
Solution

b4oody:


app/main.py
```diff
@@ -1 +1,50 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
vsmutok, line 2: `skills` variable should be an attribute of the instance not class.

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
vsmutok, line 12: Use the `extend()` method or `+=` operator to add multiple values to a single list.

vsmutok (changes requested):
image

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = FrontendDeveloper.skills + BackendDeveloper.skills
```
Dimosphen1, line 50: Do not use `__init__` in this class, rely on the implementation of the parent classes

app/main.py
```diff
@@ -1 +1,55 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
Dimosphen1, line 12: There is no need to define `skills` as a class attribute. Extend `self,skills` with required list in the `__init__` directly here and in the other classes as well

Dimosphen1 (changes requested):

vsmutok (approved):
Great!

PR #1009, opened 2024-04-25
Solution

mykola-tarasenko:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills += ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
EdAlekseiev, line 42: Why do you need this method? It looks like __init__ method of SoftwareEngineer. I think, you can delete it

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
vsmutok, line 44:
```suggestion
        self.create_powerful_api()
        self.create_awesome_web_page()
```

vsmutok (changes requested):

PR #1010, opened 2024-04-27
implement py-software-engineers

olenazaritska:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
Dimosphen1, line 43: There is no need for `__init__` method, you should rely on the implementation from the parent classes. Remove it from this class

Dimosphen1 (changes requested):

PR #1020, opened 2024-05-11
hell yeah solution

RuslanLevchuk:


app/main.py
```diff
@@ -1 +1,58 @@
...
+            "Android studio"
+        ])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 52: you don't need this, delete pls

vasylhnatiuk (changes requested):

app/main.py
RuslanLevchuk: agree

vasylhnatiuk (approved):

PR #1027, opened 2024-05-22
solution

AndrewwSSS:


app/main.py
```diff
@@ -1 +1,58 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "JavaScript",
+            "HTML",
+            "CSS"
+        ]
```
Oleg-s-s, line 18: You are redefining the list here, you should add skills to the existing list. Fix everywhere.

Oleg-s-s (changes requested):

app/main.py
AndrewwSSS: done

app/main.py
```diff
@@ -1 +1,56 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper,
+                         BackendDeveloper,
+                         SoftwareEngineer):
+    def __init__(self, name: str) -> None:
```
DevSpaciX, line 50: You do not need this method here : )

DevSpaciX (changes requested):

DevSpaciX (changes requested):
Remove redundant __init__ : )

PR #1036, opened 2024-05-27
Solution

MaksymProtsak:


app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
vasylhnatiuk, line 12: you shouldn add skills, not replace them

vasylhnatiuk: use self.skills += ["JavaScript", "HTML", "CSS"] in `init` method

vasylhnatiuk (changes requested):

app/main.py
MaksymProtsak: Hmmm.... I think, I should start from zero point.

app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.learn_skill(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def create_web_application(self) -> None:
```
vsmutok, line 46:
```suggestion
class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
    def create_web_application(self) -> None:
```

app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
```
vsmutok, line 3:
```suggestion
class SoftwareEngineer:
    def __init__(self, name: str) -> None:
```

vsmutok (approved):
Good Job!

PR #1038, opened 2024-05-27
add solution

jpdreamthug:


app/main.py
```diff
@@ -1 +1,60 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
Oleg-s-s, line 2: skills variable should be an attribute of the instance, not class

app/main.py
```diff
@@ -1 +1,60 @@
...
+        self.name = name
+
+    @classmethod
+    def learn_skill(cls, skill: str) -> None:
+        cls.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    skills = ["JavaScript", "HTML", "CSS"]
```
Oleg-s-s, line 14: And skills should be extended, that's the purpose of inheritance
fix everywhere

Oleg-s-s (changes requested):

Oleg-s-s (approved):
👍

PR #1043, opened 2024-05-28
'Solution'

dsokolovskiy:


app/main.py
```diff
@@ -1 +1,54 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        BackendDeveloper.__init__(self, name)
+        FrontendDeveloper.__init__(self, name)
+        backend_skills = BackendDeveloper(name).skills
+        frontend_skills = FrontendDeveloper(name).skills
+        self.skills = backend_skills + frontend_skills
```
DevSpaciX, line 49: You do not need it here : )

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
+        FrontendDeveloper.create_awesome_web_page(self)
```
Oleg-s-s, line 47: Wow, looks like we are doing harder here))
why not simply call methods from self?

Oleg-s-s (changes requested):

DevSpaciX (approved):

PR #1044, opened 2024-05-28
Solution

EugeneDolinskiy:


app/main.py
```diff
@@ -1 +1,54 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
pavlopro, line 44: Remove the __init__ method. It's redundant. You already inherit from BackendDeveloper and FrontendDeveloper, which have implemented all the needed logic.

app/main.py
```diff
@@ -1 +1,54 @@
...
+        BackendDeveloper.__init__(self, name)
+        FrontendDeveloper.__init__(self, name)
+        backend_skills = BackendDeveloper(name).skills
+        frontend_skills = FrontendDeveloper(name).skills
+        self.skills = backend_skills + frontend_skills
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
+        FrontendDeveloper.create_awesome_web_page(self)
```
pavlopro, line 54: Call both methods from `self`, not from classes

pavlopro (changes requested):

PR #1047, opened 2024-05-28
Solution

Andrii1000:


app/main.py
```diff
@@ -1 +1,47 @@
...
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        if isinstance(skill, str):
+            self.skills.append(skill)
+        if isinstance(skill, list):
+            self.skills = self.skills + skill
```
DevSpaciX, line 10: `self.skills.append(skill)` I think this will be enough )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1053, opened 2024-05-28
sol1

IvanVahanov:


app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = list(set(self.skills))
```
Oleg-s-s, line 43: why you did this?

Oleg-s-s (changes requested):

app/main.py
IvanVahanov: to delete person skill's copyes

DevSpaciX, line 43: You do not really need `__init__` method at all here . Just remove it

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1062, opened 2024-06-01
'Solution'

kvDevid:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        self.learn_skill("Java")
+        self.learn_skill("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 46: delete init here, you don't need it

app/main.py
```diff
@@ -1 +1,52 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
```
vasylhnatiuk, line 13: instead of duplicating use loop, or just use extend

vasylhnatiuk (changes requested):

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1067, opened 2024-06-03
'Solution'

hhorhe:


app/main.py
```diff
@@ -1 +1,53 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
vsmutok, line 16: you should not overwrite the skills, you should update them

app/main.py
```diff
@@ -1 +1,53 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = list(
+            set(FrontendDeveloper(name).skills + BackendDeveloper(name).skills)
+        )
+
```
vsmutok, line 49: This method would be redundant if you provide my changes

app/main.py
vsmutok, line 16:
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]

```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1072, opened 2024-06-11
Solution

PanValeriy:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
```
vsmutok, line 43: Remove the `__init__` method in the `FullStackDeveloper` class because it is redundant; you don't need to override `__init__` if you don't override any logic

vsmutok (changes requested):

app/main.py
PanValeriy: Thanks

vasylhnatiuk (approved):

PR #1073, opened 2024-06-12
Solution

igorchapy:


app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
vasylhnatiuk, line 33: you should add not replace skills

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = (BackendDeveloper(name).skills
```
vasylhnatiuk, line 43: if you use extend or + for skills in parents classes, you do not need these code

vasylhnatiuk (changes requested):

app/main.py
igorchapy: I don't understand what i need to fix

app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
EdAlekseiev, line 13: You should add skills instead of repalce them
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```

app/main.py
EdAlekseiev: The same comment as above

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
EdAlekseiev, line 23: You should add skills instead of repalce them
```suggestion
        self.skills += ["Python", "SQL", "Django"]
```

app/main.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

vasylhnatiuk (changes requested):
you pushed nothing

vasylhnatiuk (changes requested):
you pushed nothing

app/main.py
EdAlekseiev: Still not fixed

app/main.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #1078, opened 2024-06-20
Solution

tylerj231:


app/main.py
```diff
@@ -1 +1,63 @@
...
+
+    skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+full_stack_dev = FullStackDeveloper("Tom")
```
vasylhnatiuk, line 60: delete all code for testing

app/main.py
```diff
@@ -1 +1,63 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
vasylhnatiuk, line 12: don't replace skills but add or extend them in init

vasylhnatiuk: use += or extend()

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,57 @@
...
+
```
vsmutok, line 13: * `skills` variable should be an attribute of the instance not class.
* don't replace skills but add or extend them in init
* use `+=` or `extend()`


read [checklist.md](https://github.com/mate-academy/py-software-engineers/blob/master/checklist.md)!

vsmutok (changes requested):

app/main.py
tylerj231: Understood. Fixing it.

tylerj231: Should I do the same for Full_Stack class as well?

EdAlekseiev (approved):

PR #1081, opened 2024-06-24
Solution

Wallian169:


app/main.py
```diff
@@ -1 +1,59 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "Python",
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS"
+        ]
```
vsmutok, line 54: this is redundant here

app/main.py
```diff
@@ -1 +1,59 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```
you should not override, but increase the skils

vsmutok (changes requested):

app/main.py
Wallian169: got it

app/main.py
```diff
@@ -1 +1,51 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    BackendDeveloper,
+    FrontendDeveloper,
+    SoftwareEngineer
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Dimosphen1, line 46: There is not need for defining `__init__` here, you could rely solely on the parent classes implementation. Remove `__init__` from this class

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    BackendDeveloper,
+    FrontendDeveloper,
+    SoftwareEngineer
```
Dimosphen1, line 43: There is no need to inherit `FullStackDeveloper` since both `BackendDeveloper` and `FrontendDeveloper` inherit from it

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
Dimosphen1, line 33: Do not redefine `skills` here as well

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1083, opened 2024-06-26
Solution

juliazelena:


app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
pavlopro, line 2: skills variable should be an attribute of the instance not class.

app/main.py
```diff
@@ -1 +1,44 @@
...
+            name: str
+    ) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
```
pavlopro, line 15: You shouldn't redefine the skills attribute.
You should extend it in the __init__ method.

From the checklist:
Use the extend() method or += operator to add multiple values to a single list.

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = ["Python", "SQL", "Django"]
```
pavlopro, line 23: There same here as well

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    skills = ["Java", "Android studio"]
```
pavlopro, line 31: and here...

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
pavlopro, line 39: There is no need to define it here; it will be inherited from the parent classes

app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(
+            self,
```
pavlopro, line 5: No need to have those two parameters start on a new line
It makes sense if there are more than two

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1095, opened 2024-07-08
Solutions for software engineers

ArtemShakun:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(BackendDeveloper(name).skills)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
```
vsmutok, line 48:
```suggestion
class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
    def create_web_application(self) -> None:
        print(f"{self.name} started creating a web application...")
        self.create_powerful_api()
        self.create_awesome_web_page()
```
`__init__` method redundant here

app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]
```
you should use adding (+=) or extend() for skills instead of replacing

vsmutok (changes requested):

app/main.py
ArtemShakun: Fixed

app/main.py
ArtemShakun: Fixed

PR #1102, opened 2024-07-11
Solution

yuliiayedynak:


app/main.py
```diff
@@ -1 +1,49 @@
-# write your code here
+
+class SoftwareEngineer:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills += [skill]
```
pavlopro, line 9: Would be better to use `append `here as there is only one skill

app/main.py
```diff
@@ -1 +1,49 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
pavlopro, line 43: Do you need this `__init__`? I don't think so

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1107, opened 2024-07-15
Solution

Morphin20th:


app/main.py
```diff
@@ -1 +1,44 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"] + self.skills
```
Dimosphen1, line 13: Do not redefine `self.skills`, extend the list that is already present in `self.skills` here and in the other classes as well

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1110, opened 2024-07-19
solution

HalynaRohatska:


app/main.py
```diff
@@ -1 +1,53 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    skills = FrontendDeveloper.skills + BackendDeveloper.skills
```
MNDonut, line 48: If we have super developer who knows 10+ technologies, it means you'll need to do
skills = Dev1.skills + Dev2.skills + ... + Dev_n.skills
It's better to extend skills list inside init of each class. In this way you'll not need to do the thing above, but just inherit from classes and call super inits

MNDonut (changes requested):

PR #1113, opened 2024-07-22
Solution

DSahalatyi:


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        FrontendDeveloper.__init__(self, name)
+        frontend_skills = self.skills
+        BackendDeveloper.__init__(self, name)
+        backend_skills = self.skills
+        super().__init__(name)
+        self.skills = frontend_skills + backend_skills
```
MNDonut, line 47: it's better to do .extend() or .append() inside __init__, in this way you'll not need to add skills depending on how many classes you inherit from

MNDonut (changes requested):

PR #1116, opened 2024-07-27
solution

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1 +1,49 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = (BackendDeveloper(name).skills
+                       + FrontendDeveloper(name).skills)
```
MNDonut, line 44: 1. No need to explicitly call super because this behaviour is inherited from both classes
```
```
Also making "BackendDeveloper(name)" creates a new instance that we don't need

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1119, opened 2024-07-28
Solution

Narberal90:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        self.skills.append("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
MNDonut, line 46: no need to do it because you inherit the same behaviour from Backend and Frontend Developer

MNDonut (changes requested):

MNDonut (approved):

PR #1120, opened 2024-07-28
implement py-software-engineers

ValeriiaKyr:


app/main.py
```diff
@@ -1 +1,57 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    skills = FrontendDeveloper.skills + BackendDeveloper.skills
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
MNDonut, line 52: 1. No need to explicitly call super because this behaviour is inherited from both classes
```
```

app/main.py
```diff
@@ -1 +1,57 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
MNDonut, line 2: it's global class attribute, thus whenever you add some skills here, then all class instances will have it, but this is incorrect

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1122, opened 2024-07-29
Solution

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,57 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    base_skills = ["JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = FrontendDeveloper.base_skills
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,57 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = (
+            BackendDeveloper.base_skills
+            + FrontendDeveloper.base_skills
+        )
```
MNDonut, line 52: 1. No need to explicitly call super because this behaviour is inherited from both classes
```
```

app/main.py
```diff
@@ -1 +1,57 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    base_skills = ["JavaScript", "HTML", "CSS"]
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = FrontendDeveloper.base_skills
```
MNDonut, line 15: You'll have an issue here. Each FrontendDeveloper instance will have skills that refers to global list of skills "base_skills". And when you make self.learn_skill('new skill'), it'll add a new skill to self.skills, but self.skills is base_skills, thus each class instance will learn this skills and this is incorrect

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #1123, opened 2024-07-29
Solution for py-software-engineers

AndriiHamasa:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for skill in ["JavaScript", "HTML", "CSS"]:
+            self.learn_skill(skill)
```
LiudmylaKulinchenko, line 14: Use the extend() method or += operator to add multiple values to a single list

LiudmylaKulinchenko (changes requested):

app/main.py
AndriiHamasa: It means that we will not use learn_skill method, I thought, that if we create this method,  we should use it,
But ok.

Oleksl888 (approved):
Perfect!

PR #1127, opened 2024-07-29
Solution by Politov

Alexandr-Politov:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Oleksl888, line 42: Why do you need a super call here? It would have been required if you wanted to add something to FullStackDeveloper's `__init__` but here is unnecessary

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
```
Oleksl888, line 46: why call these methods on parent classes?
remember how does inheritance work - if FullStack inherits from both Back and Front dev which methods will he have? you can run `__dict__` on its instance to check :wink:

Oleksl888 (changes requested):

MNDonut (approved):

PR #1128, opened 2024-07-29
first_try

yashkunn:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
MNDonut, line 43: No need to call super init explicitly, python automatically does it if you don't add some custom logic

MNDonut (changes requested):

MNDonut (approved):

PR #1129, opened 2024-07-29
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        profession = ["JavaScript", "HTML", "CSS"]
+        for professions in profession:
```
Oleksl888, line 14: there must be an easier way to add a list to a list in one line without for-loop. Check and adjust in other classes

Oleksl888 (changes requested):

oliasymak24 (approved):
Good job!

PR #1130, opened 2024-07-29
first commit

panichqa:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        BackendDeveloper.__init__(self, name)
+        FrontendDeveloper.__init__(self, name)
+        self.skills = (BackendDeveloper(self.name).skills
+                       + FrontendDeveloper(self.name).skills)
```
MNDonut, line 46: This approach is quite bad because the more classes you inherit, the more such operations you should do.
Instead let's do the following: when you inherit from some class, it should automatically add its skills

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
oliasymak24, line 42: I think even no need to rewrite init method here

oliasymak24 (changes requested):

MNDonut (approved):

PR #1131, opened 2024-07-29
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = (
+            BackendDeveloper(name).skills
+            + FrontendDeveloper(name).skills
+        )
```
MNDonut, line 46: The more classes you inherit from, the more skills you'll need to set here and don't forget to add. Let's do it in another way, so that when you inherit from some class, it automatically adds its skills

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
MNDonut, line 51: no need for super(), you inherit these methods

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.add_skills(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
LiudmylaKulinchenko, line 45: What will happen if you'll remove init here?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 👏

PR #1136, opened 2024-07-30
task done

Sergi0bbb:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = list(
+            BackendDeveloper(self.name).skills
+            + FrontendDeveloper(self.name).skills
+        )
```
LiudmylaKulinchenko, line 47: The more classes you're gonna inherit from, more you're gonna have to add like this. You should try another method to add skills in classes like extending list or concatenating

LiudmylaKulinchenko (changes requested):

Oleksl888 (approved):
Well done!

PR #1137, opened 2024-07-30
Resolve

ihorutkin:


app/main.py
```diff
@@ -1 +1,54 @@
-# write your code here
+class SoftwareEngineer:
+    skills = []
```
LiudmylaKulinchenko, line 2: Why do you want it to be class attribute? What if one engineer will learn some skill, does it mean other engineers also learned it? Doesn't look logical, you should fix it

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    skills = BackendDeveloper.skills + FrontendDeveloper.skills
```
LiudmylaKulinchenko, line 46: The more classes you'll inherit from, more code you'll have to write here. Try other ways to do that

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,69 @@
...
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    # skills = BackendDeveloper.skills + FrontendDeveloper.skills
```
Oleksl888, line 41: remove unnecessary comments

app/main.py
```diff
@@ -1 +1,69 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
Oleksl888, line 13: Think how you can change this so your current developer skills will not overwrite skills of the class it is inherited from. IF you make it right it will simplify your  other logic

app/main.py
```diff
@@ -1 +1,69 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    # skills = BackendDeveloper.skills + FrontendDeveloper.skills
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        for skill in BackendDeveloper(name).skills:
```
Oleksl888, line 45: If you change the way skills are assigned in parent classes `__init__` you will not have to do this iteration

app/main.py
```diff
@@ -1 +1,69 @@
...
+            self.learn_skill(skill)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+full_stack_dev = FullStackDeveloper("Tom")
```
Oleksl888, line 54: remove testing data and comments prior pushing to remote

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1138, opened 2024-07-30
Solution

Fenya888:


vasylhnatiuk (changes requested):
try to add skills in init, and delete init in classes where it does nothing

PR #1142, opened 2024-07-31
'Solution'

RToister:


app/main.py
```diff
@@ -1 +1,66 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = (BackendDeveloper(name).skills
```
vasylhnatiuk, line 60: you don't need it, skills should be inherited from parent classes in FullStackDeveloper

vasylhnatiuk (changes requested):

PR #1155, opened 2024-08-15
Solution

razumovskayada:


app/main.py
```diff
@@ -1 +1,59 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = [
```
vasylhnatiuk, line 13: use extend() instead of replacing

app/main.py
```diff
@@ -1 +1,59 @@
...
+            "Android studio"
+        ]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 52: if you use extend in parent classes you don't need init here

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1161, opened 2024-08-19
Solution

A-n-a-s-t-a-s-i-i-a:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
vasylhnatiuk, line 13: use `extend()` instead of replacing

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = []
+        self.skills += (BackendDeveloper(self.name).skills
```
vasylhnatiuk, line 44: if you use `extend` you don't need init here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 41: you don't need init here, this class inherits skills from parents classes

vasylhnatiuk (changes requested):

PR #1165, opened 2024-08-22
My solution

lavryniuk-dev:


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
LiudmylaKulinchenko, line 47: Will it work if you'll remove init in this class? 👀

LiudmylaKulinchenko (changes requested):

app/main.py
lavryniuk-dev: Thank you, that work

LiudmylaKulinchenko (approved):
Good job 🚀

PR #1167, opened 2024-08-23
add solution for task

nikvitkaua:


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = BackendDeveloper.skills + FrontendDeveloper.skills
```
vasylhnatiuk, line 47: you don't need this if you add skills in paranet classes in init and not replace them

vasylhnatiuk (changes requested):
use extend or + to add skils in init

vasylhnatiuk (approved):

PR #1176, opened 2024-09-03
Solution

MahdaArtem-py:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
vasylhnatiuk, line 34: use += or extend

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
vasylhnatiuk, line 42: if you use use += or extend in parents classes you don't need `init` in FullStackDeveloper

vasylhnatiuk (changes requested):

app/main.py
MahdaArtem-py: sorry, but I dont understand where and why I need to use += or extend

vasylhnatiuk (approved):

PR #1188, opened 2024-09-20
Solution

KKvladys:


dmytrik (approved):
Screenshot 2024-09-21 at 19 17 01
checklist reccomend us use extend or += [skills] :) but in common i think very good

Nattalli:
> Screenshot 2024-09-21 at 19 17 01 checklist reccomend us use extend or += [skills] :) but in common i think very good

well, the checklist tells that you better add new skills (when there are a list of them) avoiding `loop` and `append` method. But in this method you are adding only one skill so you have not use a loop. So that solution is pretty good 😸

Nattalli (approved):
👍🏻

PR #1193, opened 2024-09-25
Solution

Atikiho:


app/main.py
```diff
@@ -1 +1,67 @@
...
+            "Android studio",
+        ]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
```
KKvladys, line 53: you can not use the __init__ method here if in parent classes you will use += when defining the skill attribute

KKvladys (approved):

app/main.py
```diff
@@ -1 +1,67 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
```
AnyoneClown, line 18: Use `+=` operator or `extend()` to extend a list instead of redefining it.


Check all cases in all methods

app/main.py
AnyoneClown: You can get rid of the ` __init__` method, since it inherits it from parent classes

AnyoneClown (changes requested):

PR #1196, opened 2024-09-26
created some classes and their methods

Mxbely:
Created classes: SoftwareEngineer, FrontendDeveloper, BackendDeveloper, AndroidDeveloper, FullStackDeveloper and their methods

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,54 @@
-# write your code here
+
+class SoftwareEngineer:
+    skills = []
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    skills = ["JavaScript", "HTML", "CSS"]
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    skills = ["Python", "SQL", "Django"]
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    AndroidDeveloper,
+    BackendDeveloper,
+    FrontendDeveloper,
+    SoftwareEngineer
+):
+    skills = [
+        "Python",
+        "SQL",
+        "Django",
+        "JavaScript",
+        "HTML",
+        "CSS",
+    ]
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()


app/main.py
```diff
@@ -1 +1,60 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = [
+            "Python",
+            "SQL",
+            "Django",
+            "JavaScript",
+            "HTML",
+            "CSS",
+        ]
```
Nattalli, line 55: you don't need to override that method. read checklist

Nattalli (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["JavaScript", "HTML", "CSS"])
```
Arsen-hrynevych, line 15: self.skills = ["JavaScript", "HTML", "CSS"]

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["Python", "SQL", "Django"])
```
Arsen-hrynevych, line 25: self.skills

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
```
Arsen-hrynevych, line 35: self.skills

app/main.py
Arsen-hrynevych, line 15: The reason you're directly setting self.skills to ["JavaScript", "HTML", "CSS"] in each child class (like FrontendDeveloper) instead of using self.skills.extend([...]) is primarily due to how inheritance works and the initialization of object attributes. Here's why:

In the parent class SoftwareEngineer, self.skills is initialized as an empty list:
self.skills = []

When you create an instance of FrontendDeveloper, you want to initialize it with a specific set of skills (["JavaScript", "HTML", "CSS"]). If you use self.skills.extend([...]), you are appending skills to an already initialized self.skills list. This approach might seem fine, but in certain cases, it could lead to unexpected behavior.

app/main.py
```diff
@@ -1 +1,52 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    BackendDeveloper,
+    FrontendDeveloper,
+):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Arsen-hrynevych, line 47: You are inheriting from both BackendDeveloper and FrontendDeveloper. Since both classes are children of SoftwareEngineer, Python's method resolution order (MRO) will only call the __init__ method from the first parent class (BackendDeveloper).
This means the FrontendDeveloper's skills won't be initialized unless you call the __init__ of both parents explicitly.

Arsen-hrynevych (changes requested):
Mostly correct

app/main.py
Mxbely: I understand it, and know about MRO, and i did it first

app/main.py
Mxbely: I understand about MRO and about inherited skills, but in the example don't say anything about FullStackDeveloper skills , just have an example with they skills, and i don't known, we need inherited skills from Beckend and Frontend classes or override skills?
in the first time i do this:
```
        super().__init__(name)
        self.skills = [
            "Python",
            "SQL",
            "Django",
            "JavaScript",
            "HTML",
            "CSS",
        ]
```
but Natalia made a reject
i make it through ....extend.... and also have a reject
so, what i am doing wrong?

if i doing smth like that:... i extended Backend skills, and if i understand, i haven't do it
```
class FullStackDeveloper(
    BackendDeveloper,
    FrontendDeveloper,
):
        super().__init__(name)
        self.skills.extend([
            "JavaScript",
            "HTML",
            "CSS"
        ])
```

Nothing about init skills:       and my first code isn't correct, and we shouldn't extend parent classes skills, why?
image

Mxbely: I think, my solve is correct

app/main.py
Mxbely: I think, my solve is correct

app/main.py
Nattalli: Hello, you’re right that it’s not directly asked in the task, but in general, for a full-stack developer, it is assumed that there’s no need to explicitly list of skills, as it should be a union of both backend and frontend skills.

app/main.py
```diff
@@ -1 +1,57 @@
...
+        self.skills.extend([
+            "JavaScript",
+            "HTML",
+            "CSS"
+        ])
```
Nattalli, line 52: so code will work without that method

Nattalli: if you are interested - add prints

Nattalli (changes requested):

Nattalli (approved):

PR #1199, opened 2024-10-01
Solution

VladyslavBon:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        self.learn_skill("Android studio")
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
vladyslav-tmf, line 47: The code is well-structured, but there's no need to override the **__init__** method in the **FullStackDeveloper** class since all skills are combined through the inheritance order from **BackendDeveloper** and **FrontendDeveloper**. This makes the constructor unnecessary, as the skills will automatically include those from both parent classes.

vladyslav-tmf (changes requested):

vladyslav-tmf (approved):

app/main.py
```diff
@@ -1 +1,49 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill("JavaScript")
+        self.learn_skill("HTML")
+        self.learn_skill("CSS")
```
Nattalli, line 15: can you extend `skills` list in one line? calling only one method (you can see at checklist)

Nattalli: the same everywhere

Nattalli (changes requested):

Nattalli (approved):

PR #1200, opened 2024-10-01
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -1 +1,45 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills: list[str] = []
+
+    def learn_skill(self, skill: str) -> None:
+        if skill not in self.skills:
+            self.skills.append(skill)
```
Mxbely, line 8: you can not use `if` in `learn_skill` method, if you have created list in the `__init__ `method, because you already have this list

Mxbely (changes requested):

app/main.py
vladyslav-tmf: Method **__init__** creates an empty list for storing skills.
The **learn_skill** method adds a new skill, but first checks if the list already contains this skill to avoid duplicates.
So, in my opinion, the verification is important.

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["JavaScript", "HTML", "CSS"])
```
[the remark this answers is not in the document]
vladyslav-tmf: The **self.skills** attribute is already initialized as an empty list in the parent **SoftwareEngineer** class when the **__init__** method is called. In child classes, such as FrontendDeveloper, there is no need to create it again.
The use of `self.skills.extend(...`) in child classes is correct because it simply extends the existing list of skills inherited from the parent class.

Nattalli (approved):

PR #1206, opened 2024-10-04
'Solution'

spa7id:


skyfoxwork (approved):
I think good job!

But if you will use:

    self.skills.extend([skill])

instead of:

    self.skills.append(skill)

and 
 
insted of:

    self.skills = ["JavaScript", "HTML", "CSS"]

and so on it other clesses.

I think it will better!

And then you don't need __init__() method in FullStackDeveloper class.

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL",
+                       "Django", "JavaScript",
+                       "HTML", "CSS"]
```
skyfoxwork, line 45: if you will use self.skills.extend(["skill_one", "skill_two", "skill_three"]) you don't need __init__ () method.

app/main.py
```diff
@@ -1 +1,50 @@
...
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
```
skyfoxwork, line 41: if you will use self.skills.extend(["skill_one", "skill_two", "skill_three"]) you don't need __init__ () method.

app/main.py
```diff
@@ -1 +1,50 @@
...
+        super().__init__(name)
```
Arsen-hrynevych, line 42: You are inheriting from both BackendDeveloper and FrontendDeveloper. Since both classes are children of SoftwareEngineer, Python's method resolution order (MRO) will only call the __init__ method from the first parent class (BackendDeveloper).
This means the FrontendDeveloper's skills won't be initialized unless you call the __init__ of both parents explicitly.

Arsen-hrynevych (approved):
Mostly correct

app/main.py
skyfoxwork: I agree with you!

But it work without 

    __init__()   method
    
I think it is because when we use method self.create_awesome_web_page() that inherited from Frontend Developer it will run __init__() method from FrontendDeveloper.

PR #1208, opened 2024-10-05
Solution-py-software-engineers

skyfoxwork:


Andrey-Dementyev (approved):

app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.extend([skill])
```
Nattalli, line 7: just `append` would be better here as you need add only one skill

Nattalli (changes requested):

Nattalli (approved):

PR #1209, opened 2024-10-05
'Solution'

Andrey-Dementyev:


app/main.py
```diff
@@ -1 +1,49 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django", "JavaScript", "HTML", "CSS"]
```
Nattalli, line 44: read checklist, you don't need to overwrite that method if it's already inplemented in parent classes (but you should provide adding skills there) and fullstack developer skills is frontend developer skills + backend developer skills

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
Nattalli, line 33: that should extend list of skills from `SoftwareEngineer` class

Nattalli (changes requested):

PR #1210, opened 2024-10-06
Solution

anomelchuk:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
```suggestion
        self.skills += ["JavaScript", "HTML", "CSS"]

```

vasylhnatiuk: if you use adding instead of replacing in parent classes you don't need init in FullStackDeveloper

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1215, opened 2024-10-07
solution

pohgen:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["JavaScript", "HTML", "CSS"])
```
mshmygel, line 13: can You use learn_skill() here?

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
mshmygel, line 47: “self” would be better to use instead of super()

mshmygel (approved):

app/main.py
Nattalli: actually `extend` is good solution

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
Nattalli, line 42: read checklist, you don't need to overwrite that method if it's already inplemented in parent classes

Nattalli (changes requested):

PR #1217, opened 2024-10-07
Solution

molodsh1y1:


sumseer (approved):
gj

app/main.py
```diff
@@ -1 +1,51 @@
...
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name=name)
+        self.skills = [
+            "Java",
+            "Android studio",
+        ]
```
Arsen-hrynevych, line 40: If you previously used self.learn_skill() to add skills, but now directly modify self.skills, it creates inconsistency.

Arsen-hrynevych (approved):
Ok

app/main.py
molodsh1y1: Ok

PR #1221, opened 2024-10-08
solution

VolodymyrSemchysyn:


app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = (FrontendDeveloper(name).skills
```
Arsen-hrynevych, line 43: the FullStackDeveloper class is attempting to manually combine the skills of FrontendDeveloper and BackendDeveloper by creating new instances of these classes within the constructor:

self.skills = (FrontendDeveloper(name).skills + BackendDeveloper(name).skills)
Problem: This approach creates new instances of FrontendDeveloper and BackendDeveloper, which leads to redundant objects that have no connection to the current instance of FullStackDeveloper.

Impact: These redundant instances will not inherit any skills that might have been learned by the FullStackDeveloper instance through its learn_skill method, as they are isolated objects.

Fix: Instead of instantiating new FrontendDeveloper and BackendDeveloper objects, you should let the multiple inheritance mechanism work as intended, allowing FullStackDeveloper to inherit the behavior and attributes directly from its parent classes. In the corrected code, the __init__ method of FullStackDeveloper is not needed, and the skills attribute can be inherited naturally through multiple inheritance.

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
Arsen-hrynevych, line 23: each subclass (FrontendDeveloper, BackendDeveloper, AndroidDeveloper) assigns a new list of skills directly to self.skills:

self.skills = ["JavaScript", "HTML", "CSS"]

Problem: This overrides the skills list that might have been previously populated (e.g., via the learn_skill method in the base SoftwareEngineer class), meaning any learned skills will be lost.

Fix: The correct way is to extend the existing skills list using extend():

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
```
Arsen-hrynevych, line 33: each subclass (FrontendDeveloper, BackendDeveloper, AndroidDeveloper) assigns a new list of skills directly to self.skills:

self.skills = ["JavaScript", "HTML", "CSS"]

Problem: This overrides the skills list that might have been previously populated (e.g., via the learn_skill method in the base SoftwareEngineer class), meaning any learned skills will be lost.

Fix: The correct way is to extend the existing skills list using extend():

app/main.py
```diff
@@ -1 +1,49 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
Arsen-hrynevych, line 13: each subclass (FrontendDeveloper, BackendDeveloper, AndroidDeveloper) assigns a new list of skills directly to self.skills:

self.skills = ["JavaScript", "HTML", "CSS"]

Problem: This overrides the skills list that might have been previously populated (e.g., via the learn_skill method in the base SoftwareEngineer class), meaning any learned skills will be lost.

Fix: The correct way is to extend the existing skills list using extend():

Arsen-hrynevych (changes requested):
Please review comments above

Arsen-hrynevych (approved):
ok

PR #1312, opened 2024-12-05
Software Engineer constructor

valerii-kashpur:
- added SoftwareEngineer class
- added FrontendDeveloper class
- added BackendDeveloper class
- added AndroidDeveloper class
- added FullStackDeveloper class

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,48 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Java", "Android studio"]
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["JavaScript", "HTML", "CSS"]
```
L1nk3rrr, line 13: Don't need to reassign the list, you already have a list on `self.skills`!

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills = ["Python", "SQL", "Django"]
```
L1nk3rrr, line 23: Don't need to reassign the list, you already have a list on `self.skills`!

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills += BackendDeveloper(name).skills
```
L1nk3rrr, line 43: You can avoid this overwriting of `__init__`, if you will extend skills at `FrontendDeveloper`, `BackendDeveloper`

L1nk3rrr (changes requested):
Clean code, but try to improve it with my comments! Use the full power of inheritance!

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
L1nk3rrr, line 42: you don't need this __init__ overriding as you simply call super

L1nk3rrr: try to run your code without this and figure out why this is working

L1nk3rrr, line 42: you don't need this __init__ overriding as you simply call super
try to run your code without this and figure out why this is working

L1nk3rrr (changes requested):

app/main.py
valerii-kashpur: will be called init from parents (FrontendDeveloper, BackendDeveloper)

L1nk3rrr (approved):
Well done! Great work!

PR #1337, opened 2024-12-10
Solution

Arsenmyron:


app/main.py
```diff
@@ -1 +1,50 @@
...
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        if isinstance(skill, str):
+            self.skills.append(skill)
+        else:
+            self.skills.extend(skill)
```
Dogober, line 10: Here is no need to do any check. This method needs only for adding skill(str) especially for this class

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill(["JavaScript", "HTML", "CSS"])
```
Dogober, line 16: Here you can remove a calling learn_skill func and using extend method for adding the skills as example:

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_awesome_web_page(self) -> str:
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill(["Python", "SQL", "Django"])
```
Dogober, line 26: as well as in previous class

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.learn_skill(["Java", "Android studio"])
```
Dogober, line 36: as well as in previous class

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        BackendDeveloper.create_powerful_api(self)
+        FrontendDeveloper.create_awesome_web_page(self)
```
Dogober, line 50: If you inherit from FrontendDeveloper and BackendDeveloper classes you can use self.create_powerful_api instead of BackendDeveloper.create_powerful_api(self)

Dogober (approved):

PR #1338, opened 2024-12-10
'Solution'

Dogober:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
+    BackendDeveloper,
+    FrontendDeveloper,
+    SoftwareEngineer
+):
+    def __init__(self, name: str) -> None:
```
Sichkaridze, line 45: You don't need this/
But still great.

Sichkaridze (approved):

app/main.py
Dogober: Yep I got it Thx

PR #1379, opened 2025-01-01
implement SoftwareEngineer, FrontendDeveloper, BackendDeveloper 

yarzhuravlov:
and FullStackDeveloper classes

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,47 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills: list[str] = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
oliasymak24, line 42: No need to rewrite init method here

oliasymak24 (changes requested):

app/main.py
yarzhuravlov: Changes applied, thanks for the feedback

oliasymak24 (approved):

PR #1476, opened 2025-02-24
Solution

d-stegantsev:


Sotnikov-100 (approved):
If you add self.skills = list(set(self.skills)) to the initializer of the FullStackDeveloper class, it will help avoid duplicates and make your code more refined. Try it in the test. Good Job🥳🥳

PR #1481, opened 2025-02-25
Solution software_engineers

Roksolana-K:


app/main.py
```diff
@@ -1 +1,56 @@
...
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(
```
Claymore-AI, line 44: You should not include SoftwareEngineer here, as it is the parent class  for FrontendDeveloper and BackendDeveloper. Including it would be redundant because they already inherit the functionality from SoftwareEngineer.

PR #1491, opened 2025-02-26
task solution

yarms1:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        SoftwareEngineer.__init__(self, name)
+        backend_master = BackendDeveloper(name).skills
+        frontend_master = FrontendDeveloper(name).skills
+        self.skills = sorted(list(set(backend_master + frontend_master)))
```
Sezaro104, line 45: Use super() instead of the current construct. super() will allow you to immediately inherit the addition of elements to the list, and will shorten the entire construction.

Sezaro104 (approved):

PR #1497, opened 2025-03-01
Solution

BuryiOstap:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
vsmutok, line 42: `__init__ ` method redundant in this case

BuryiOstap: because this method is already implemented in parent functions?

BuryiOstap: because this func already implemented in parent classes?

vsmutok (approved):
Looks Good!

PR #1543, opened 2025-03-27
Implement FullStackDeveloper with combined skills.

farzamgt:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["JavaScript", "HTML", "CSS"])
```
Tolik-code, line 13: In parent class SoftwareEngineer you have learn_skill method, use it instead of setting directly skills

Tolik-code (approved):
One comment above, but everythings good!

MNDonut (approved):

PR #1551, opened 2025-04-03
Solution

Michal-Naron:


Michal-Naron:
@mateacademy-ai-mentor tag3.
Check this one more time I even give you a comment in a new commit I use self. to turn on method form paret class

PR #1602, opened 2025-05-07
Solution

Nib20:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    def learn_skill(self, skill: str) -> None:
+        self.skills.append(skill)
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.append("JavaScript")
+        self.skills.append("HTML")
+        self.skills.append("CSS")
```
MNDonut, line 15: let's use `extend` here to not overwhelm the code

app/main.py
```diff
@@ -1 +1,52 @@
...
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.append("Python")
+        self.skills.append("SQL")
+        self.skills.append("Django")
```
MNDonut, line 27: the same here

app/main.py
```diff
@@ -1 +1,52 @@
...
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.append("Java")
+        self.skills.append("Android studio")
```
MNDonut, line 38: and here

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
MNDonut, line 47: by default if you don't define some method, python will use MRO to find the nearest init and will call its implementation, so no need to create the empty method

MNDonut (changes requested):

app/main.py
Nib20: fixed

app/main.py
Nib20: fixed

app/main.py
Nib20: fixed

app/main.py
```diff
@@ -34,17 +30,15 @@ def create_powerful_api(self) -> str:
...
 
     def create_smooth_mobile_app(self) -> str:
         print(f"{self.name} is creating a mobile app...")
         return "Ads every three swipes"
 
 
 class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
-    def __init__(self, name: str) -> None:
-        super().__init__(name)
+    pass
```
MNDonut, line 41: actually there shoudn't be `pass` because your class is not fully empty

MNDonut (changes requested):

app/main.py
Nib20: fixed

MNDonut (approved):

PR #1662, opened 2025-06-22
Solution

DrGregoryHouse4676:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
MNDonut, line 42: no need to call super init explicitly because it'll called automatically by MRO when it isn't found in the current class

MNDonut (changes requested):

MNDonut (approved):

PR #1668, opened 2025-06-26
my-solution

Orixara:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
MNDonut, line 42: No need to call init explicitly here because if python doesn't find init here, it'll go by MRO

MNDonut (changes requested):

PR #1681, opened 2025-06-29
Solution for Software engineers

yagginov:


Vskesha (approved):
Good working solution of the task.

PR #1682, opened 2025-06-29
Solution

Vskesha:


app/main.py
```diff
@@ -1 +1,48 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
```
dimashevchukk, line 4: adding list[str] annotation would be good

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        self.create_powerful_api()
+        self.create_awesome_web_page()
+
+
+full_stack_dev = FullStackDeveloper("Tom")
+print(full_stack_dev.skills)
```
dimashevchukk, line 48: forgot to remove testing lines :)

dimashevchukk (approved):

app/main.py
Vskesha: fixed

app/main.py
Vskesha: thanks, fixed

Vskesha:
@mateacademy-ai-mentor we are going to call the init method along the mro chain and it will be like this in this case:
FullStackDeveloper ->FrontendDeveloper -> BackendDeveloper -> SoftwareEngineer -> object
Therefore, both backend and frontend skills will be added. super() in this case will not point to the parent class in __init__ of the frontend, but to the next class in mro.

Also, the condition for the fullstack developer class does not specify that its own __init__ is needed

PR #1686, opened 2025-06-29
Solution

Kostenikov:


Kostenikov:
@mateacademy-ai-mentor we are going to call the **init** method along the mro chain and it will be like this in this case: FullStackDeveloper ->FrontendDeveloper -> BackendDeveloper -> SoftwareEngineer -> object 
Therefore, both backend and frontend skills will be added. `super()` in this case will not point to the parent class in `__init__` of the frontend, but to the next class in mro.

Also, the condition for the fullstack developer class does not specify that its own `__init__` is needed

PR #1692, opened 2025-06-30
solution

MorsImmortalis05:


app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
L1nk3rrr, line 52: Looks like no need to overwrite __init__

L1nk3rrr (changes requested):
Try to test FullStack without overriding __init__ -> result will be the same

PR #1701, opened 2025-07-01
created SoftwareDeveloper class with subclasses

trellle:


bohdankolisnyk04 (approved):

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.skills.extend(["Java", "Android studio"])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(FrontendDeveloper, BackendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
L1nk3rrr, line 42: no sense to override, as you didnt make any changes. You can remove overriding of init

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def create_powerful_api(self) -> str:
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        self.skills.extend(["Java", "Android studio"])
```
L1nk3rrr, line 33: You have realized `learn_skill` but never used it. Also every time extending of list looks like code duplication, maybe better to make a method that will extend skills?
Smth like learn_skils that will accept a list, and then it will extend self.skills

L1nk3rrr (changes requested):
Try to improve your code!

PR #1924, opened 2025-12-11
Solution

herasymchukkdenysss:


herasymchukkdenysss:
it works correctly after calling frontenddevelop __init__ it call backenddevelop __init__ and so on. in mro order, I even did a print debug ['Python', 'SQL', 'Django', 'JavaScript', 'HTML', 'CSS'] here is the result, you're wrong

PR #1961, opened 2026-01-06
added code to cover requirements

yuliiazahorulko:


yuliiazahorulko:
@mateacademy-ai-mentor 

Please review again, there was no such defect.
You are wrong. 

If you don't agree that I wrote

print(f"{self.name} started creating a web application...")

Please give me a line or my code snippet that should be updated.

Thanks!

PR #2026, opened 2026-02-09
Solution: implemented classes to complete this task

SerhiiHryniuk:


app/main.py
```diff
@@ -1 +1,50 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
+
+    def learn_skill(self, skill: str) -> None:
+        self.skills += [skill]
```
Reverad: it's working but looks weird, ur creating list with one str element for extend() yr self.skills, when u could just .append() this str

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+class FrontendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        skills = ["JavaScript", "HTML", "CSS"]
+        for skill in skills:
+            super().learn_skill(skill)
```
Reverad: exactly here u could use .extend() method for extend yr list(self.skills) which yr creating on line 12

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print(f"{self.name} is creating a webpage...")
+        return "<h1>Hello world</h1>"
+
+
+class BackendDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        skills = ["Python", "SQL", "Django"]
+        for skill in skills:
+            super().learn_skill(skill)
```
Reverad: the same problem as in the previous comm

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print(f"{self.name} is creating an API...")
+        return "http://127.0.0.1:8000"
+
+
+class AndroidDeveloper(SoftwareEngineer):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
+        skills = ["Java", "Android studio"]
+        for skill in skills:
+            super().learn_skill(skill)
```
Reverad: the same to previous comm

app/main.py
```diff
@@ -1 +1,50 @@
-# write your code here
+class SoftwareEngineer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.skills = []
```
Reverad: u could add type annotation here "list[str]"

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def create_web_application(self) -> None:
+        print(f"{self.name} started creating a web application...")
+        super().create_powerful_api()
+        super().create_awesome_web_page()
```
Reverad: it's worth adding an explicit __init__ here. currently the class relies on mro and the __init__ of the first parent(BackendDeveloper), this makes code fragile: if the order of inheritance or the structure of the parent constructors changes, the initialization will break. 
you should use self instead of super() here, using super() explicitly tells python to skip the current class and look for the method in the parent classes, it's better to use self so that if you ever override these methods your code will correctly use the new version

Reverad (changes requested):

Reverad (approved):
that's it, gj!!

PR #2069, opened 2026-03-04
Solution

Reverad:


app/main.py
```diff
@@ -1 +1,58 @@
...
+        ])
+
+    def create_smooth_mobile_app(self) -> str:
+        print(f"{self.name} is creating a mobile app...")
+        return "Ads every three swipes"
+
+
+class FullStackDeveloper(BackendDeveloper, FrontendDeveloper):
+    def __init__(self, name: str) -> None:
+        super().__init__(name)
```
pawunder1992: You can delete this block of code according to checklist "Avoid overriding __init__() if you're only calling super().__init__() with the same arguments" , it is dont needed.

pawunder1992 (approved):
You can change .append to +=[] as wrote in checklist, you could used another logic without super but it works.GJ

PR #2119, opened 2026-03-25
very hard solution

pawunder1992:


lubkoserhii (approved):
There's an issue with the skills. It would be better to add the skills to the __init__ method of the parent class `SoftwareEngineer`, then add them to `learn_skill` using the `append` method. After that, add them to the child classes using the `extend` method.

PR #2143, opened 2026-04-03
Solution

Sacchar20:


andriihrechko (approved):

alisahoys (approved):
Needs composition thinking for a real-world readiness, but inheritance is written in a correct OOP way

alisahoys (approved):

